PR #9, opened 2016-01-05
Add more ignored files and folders

rastersize:
This PR adds more files and “folders” that git should ignore. So that we don’t have to depend on users having a global ignore file.

@8W9aG


8W9aG:
:+1:

PR #10, opened 2016-01-05
Use xcpretty to format output when building and testing on Travis-CI

rastersize:
Should result in output which is quite a bit easier to read. The PR also makes sure the correct locale is set.

Also opt-in to not using `sudo` this also opts us in to using the newer Travis-CI architecture.


8W9aG:
:+1:

PR #11, opened 2016-01-05
Remove second slather call

rastersize:
The `coverage` call will also upload it.

PR #17, opened 2016-01-12
Simplify slather + coveralls

dflems:
We were calling slather twice. The second time, ignore wasn't set properly, so system framework coverage data was being posted to coveralls. This PR *should* simplify things. If this works, we can close #11.

### DIFF
--- a/ci/after_success.sh
+++ b/ci/after_success.sh
@@ -4,10 +4,5 @@ set -euo pipefail
 
 bundle exec slather coverage \
     --input-format profdata \
-    --cobertura-xml \
-    --ignore "../**/*/Xcode*" \
-    --output-directory slather-report \
     --scheme SPTDataLoader \
     SPTDataLoader.xcodeproj
-
-bundle exec slather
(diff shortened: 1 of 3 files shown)


dflems:
@jgavris @rastersize: seems to have worked ;)

dflems:
Note that coverage dropped because I excluded the files in `SPTDataLoaderTests/*`. A more accurate number is now reflected by the coverage report.

jgavris:
Thanks @dflems! I'll bump the coverage limit back to 78%.

ci/after_success.sh
```diff
@@ -4,10 +4,5 @@ set -euo pipefail
 
 bundle exec slather coverage \
     --input-format profdata \
-    --cobertura-xml \
-    --ignore "../**/*/Xcode*" \
-    --output-directory slather-report \
```
jgavris: We can remove this from the `.gitignore` now as well?

dflems: Yup. I'll delete that and test that the new limit is working.

rastersize:
Awesome, thanks @dflems!

jgavris:
:+1: Ignore the coverage decrease...

PR #18, opened 2016-01-12
Make -dataLoader:didCancelRequest: an optional delegate method

JohnSundell:
Most cancellations are done by the API user, so this notification is often not used to implement any logic, and has most often be left empty since it’s been a required method.

This PR makes it optional, so that API users that are interested in this event can react to it, while still removing boilerplate for the most common use case.


JohnSundell:
@8W9aG

8W9aG:
:+1:

rastersize:
that is awesome

PR #24, opened 2016-01-13
Support OSX

dflems:
Experimental. Let's see if this works.

@jgavris @8W9aG @rastersize

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -3,6 +3,7 @@
 set -euo pipefail
 
 xcrun xcodebuild build test \
+	NSUnbufferedIO=YES \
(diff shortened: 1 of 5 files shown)


jgavris:
:+1:

ci/script.sh
```diff
@@ -3,11 +3,12 @@
 set -euo pipefail
 
 xcrun xcodebuild build test \
+	NSUnbufferedIO=YES \
```
jgavris: tabs / spaces :grin:

dflems:
@jgavris: I think we might be going a little overkill with all those builds ;)

PR #25, opened 2016-01-13
Update Demo

dflems:
* Moved demo code to new target in main project
* Fix broken positioning in main storyboard
* Normalize location of client key & secret
* Update README with instructions on how to use it

@jgavris @8W9aG @rastersize


8W9aG:
:+1:

PR #28, opened 2016-01-13
Fix platform/arch config

dflems:
Running tests from the Xcode IDE was broken (though xcodebuild still worked).

@8W9aG @rastersize @jgavris


8W9aG:
:+1:

PR #30, opened 2016-01-14
Add deploy ci

8W9aG:


8W9aG:
@dflems @jgavris @rastersize

.travis.yml
```diff
@@ -39,3 +39,12 @@ before_install: ./ci/before_install.sh
...
 script: ./ci/script.sh
 after_success: ./ci/after_success.sh
+deploy:
+  provider: releases
+  api_key:
+    secure: ANvlZh4vgRuTeUE9IenqsiKU0iqHVVJGDmqjvqmg0PMcN6QgR7leAAf45cQvPwKeJqS7DeXHFLq7gs0c7gfofBluFpt/UdCcHb1L7aXG+14uPVp6f5ggK5ZUxXwWfkW4s1xtAvDvQmGo7qB9Q/EaUSHJvH2vOBPlaaFc+h6C7M4=
+  file:
+    SPTDataLoader.framework.zip
+  on:
+    repo: spotify/SPTDataLoader
```
jgavris: `tags: true` or `branch: master`?

rastersize: I’d say `tags: true`.

As we should only make binary releases for our tags IMO.

Gemfile
```diff
@@ -6,3 +6,4 @@ gem 'slather',
   :branch   => "feature-profdata"
 gem 'xcpretty'
 gem 'xcpretty-travis-formatter'
+gem 'carthage'
```
rastersize: Don’t think there’s a carthage gem? You install it via Homebrew. Also Travis-CI has it installed by default.

We might want to make sure it’s the latest though, so in our `ci/before_install.sh` script add:
```shell
brew update
brew install carthage
```

jgavris: I'd probably move it into `deploy.sh` as that's the only consumer.

jgavris: You need to check in any changes to a gemfile with the gemfile.lock. I'd just install this gem afterwards in `deploy.sh` if all goes well prior to that step.

.travis.yml
8W9aG: Yeh will make it tags: true

Gemfile
8W9aG: Will move into deploy.sh

jgavris:
:+1:

rastersize:
:tada: :ship: :it:

PR #33, opened 2016-01-21
Move languageHeaderValue to the public API

8W9aG:
* We need to get this so our C++ core can append the
same Accept-Language header to our proprietary
transport layer


8W9aG:
@dflems @rastersize @jgavris

jgavris:
:+1:

rastersize:
:+1:

PR #34, opened 2016-01-21
Move tags: true to the on: section in travis

8W9aG:
* Made a mistake when putting in the root of deploy


8W9aG:
@jgavris @rastersize @dflems

jgavris:
:+1:

PR #38, opened 2016-01-22
Use incredibly strict warning set for SPTDataLoader

8W9aG:
* Removes the use of spotify_shared.xcconfig
* Disables a handful of warnings, but generally uses -Weverything


8W9aG:
@rastersize @dflems @jgavris @JensAyton

8W9aG:
From Spotify's perspective are we just using spotify_shared for common warnings on our modules? Is there anything else that needs to be considered?

jgavris:
:+1:

8W9aG:
Will merge this now, will revert if it causes troubles with our build system

PR #39, opened 2016-01-22
Fix the remaining issues with no spotify_shared.xcconfig

8W9aG:
* Add some more options to spotify_os.xcconfig
* Make the target inherit these settings
* Fix some more test warnings
* Add a way to allow all certificates through code rather than macros

### DIFF
--- a/include/SPTDataLoader/SPTDataLoaderService.h
+++ b/include/SPTDataLoader/SPTDataLoaderService.h
@@ -37,7 +37,7 @@
  * builds.
  * @warning This will trigger an assert if all certificates are allowed on release builds.
  */
-@property (nonatomic, assign, readwrite, getter = isAllCertificatesAllowed) BOOL allCertificatesAllowed;
+@property (nonatomic, assign, readwrite, getter = areAllCertificatesAllowed) BOOL allCertificatesAllowed;
 
 /**
  * Class constructor
--- a/include/SPTDataLoader/SPTDataLoaderService.h
+++ b/include/SPTDataLoader/SPTDataLoaderService.h
@@ -32,6 +32,14 @@
 @interface SPTDataLoaderService : NSObject
 
 /**
+ * Whether all certificates are allowed (e.g. broken certificates don't cause closed connections)
+ * @discussion By default this is NO. You should only turn this on if you are certain it will not be on in release
+ * builds.
+ * @warning This will trigger an assert if all certificates are allowed on release builds.
+ */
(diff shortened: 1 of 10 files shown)


8W9aG:
@dflems @rastersize @jgavris @JensAyton

include/SPTDataLoader/SPTDataLoaderService.h
```diff
@@ -32,6 +32,14 @@
 @interface SPTDataLoaderService : NSObject
 
 /**
+ * Whether all certificates are allowed (e.g. broken certificates don't cause closed connections)
+ * @discussion By default this is NO. You should only turn this on if you are certain it will not be on in release
+ * builds.
+ * @warning This will trigger an assert if all certificates are allowed on release builds.
+ */
+@property (nonatomic, assign, readwrite, getter = isAllCertificatesAllowed) BOOL allCertificatesAllowed;
```
jgavris: `areAllCertificatesAllowed`

spotify_os.xcconfig
```diff
@@ -10,6 +10,8 @@ ENABLE_NS_ASSERTIONS_Debug = YES
 ENABLE_NS_ASSERTIONS_Release = NO
 ONLY_ACTIVE_ARCH_Debug = YES
 ONLY_ACTIVE_ARCH_Release = NO
+ONLY_ACTIVE_ARCH = $(ONLY_ACTIVE_ARCH_$(CONFIGURATION))
+SUPPORTED_PLATFORMS = macosx iphoneos iphonesimulator watchsimulator tvsimulator
```
rastersize: I think it might be good to add `CLANG_MODULES_AUTOLINK = NO` as well. So that we don’t add dependencies on other modules accidentally. I.e. making sure it requires thought.

rastersize:
I re-triggered the failed build. Looked like the simulator flaking out.

However @8W9aG seems like you’ll need to add some more unit tests :smile: 

---

Did notice that Coveralls isn’t able to find the source files from our repo. Not sure why though. See https://coveralls.io/builds/4818416/source?filename=SPTDataLoader%2FSPTDataLoaderService.m.

include/SPTDataLoader/SPTDataLoaderService.h
8W9aG: :+1: Will make the change

spotify_os.xcconfig
8W9aG: :+1: Will make teh change

8W9aG:
@JensAyton

JensAyton:
:+1:

PR #45, opened 2016-01-26
Add tvOS and watchOS to podspec

rastersize:
`master` supports tvOS and watchOS so let’s declare it in our podspec.

Also fix wrong type of apostrophe in the description.

@8W9aG @jgavris @dflems

### DIFF
--- a/SPTDataLoader.podspec
+++ b/SPTDataLoader.podspec
@@ -5,7 +5,7 @@ Pod::Spec.new do |s|
     s.summary      = "SPTDataLoader is Spotify's HTTP library for Objective-C"
 
     s.description  = <<-DESC
-                        Authentication and back-off logic is a pain, let's do it
+                        Authentication and back-off logic is a pain, let’s do it
                         once and forget about it! This is a library that allows you
                         to centralise this logic and forget about the ugly parts of
                         making HTTP requests.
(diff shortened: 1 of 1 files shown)


8W9aG:
:+1:

SPTDataLoader.podspec
```diff
@@ -13,6 +13,8 @@ Pod::Spec.new do |s|
 
     s.ios.deployment_target = "6.0"
     s.osx.deployment_target = "10.8"
+    s.tvos.deployment_target = '9.0'
+    s.watchos.deployment_target = '2.0'
```
jgavris: Have wanted to fix cocoapods to introspect this from the project file :)

jgavris:
:+1:

SPTDataLoader.podspec
rastersize: @jgavris That would be a great enhancement to CocoaPods, indeed :)

rastersize:
Hmm, Travis-CI does not want to build this PR for some reason:
screen shot 2016-01-26 at 18 10 23

See the [requests page for SPTDataLoader](https://travis-ci.org/spotify/SPTDataLoader/requests).

SPTDataLoader.podspec
dflems: AFAIK, CocoaPods doesn't do any introspection from the project file. Some pods have no project file at all. I kind of like that everything has to be explicit in the podspec especially as the pod target is generated procedurally rather than copying over settings from the weirdness that is xcode projects.

PR #46, opened 2016-01-26
Add a build step which validates each source file for the inclusion of our license header

rastersize:
This PR adds a build step which validates that all source files contains our license header. The script did find a few files which where missing it :smile: 

It also adds Travis-CI group folding to our different steps, to make the output pretties and easier to find. 

Credits for the nice little `diff` command goes to @dflems :tada:


ci/script.sh
```diff
@@ -1,14 +1,29 @@
...
     -sdk "$TEST_SDK" \
     -destination "$TEST_DEST" \
     $EXTRA_ARGUMENTS \
-         | xcpretty -c -f `xcpretty-travis-formatter`
+        | xcpretty -c -f `xcpretty-travis-formatter`
+
+
+# License conformance
+travis_fold_open "License conformance"
+./ci/validate_license_conformance.sh {include/SPTDataLoader/*.h,SPTDataLoader/*.{h,m}}
```
rastersize: I put the license validation after the general build actions. However as they run so fast, and they will fail the build, should we have the check before building and test? Wondering if that would make the fail/fix cycle quicker.

Downside to that is that you’d have to fix the license issue (trivial) before you’re even told about potential code failures.

8W9aG:
I'm for putting it before the build, it seems like it is better to fail fast

rastersize:
@8W9aG Ok, I moved it.

rastersize:
This is ready for review @8W9aG @jgavris @dflems :smiley:

jgavris:
Awesome efforts! I am mostly against merging 'revert' and 'fixup' commits into mainline from a topic branch. How do others feel about squashing into the winner commits?

8W9aG:
I think each to their own on this issue, their are valid arguments for both

rastersize:
@jgavris @8W9aG @dflems Ready for review!

@jgavris I went ahead and squashed the commits a bit. It should be less noisy now. However I left the commits which do separate things separate. Also some commits would have been annoying to squash (specifically squashing 5d0cafa4707c84949fb2bfe1a4faaffc04f1a83f into 838963f9307facb417963ec82d376c574c218885) so I left those as-is.

8W9aG:
:+1: Very nice

8W9aG:
:+1:

jgavris:
:+1:

PR #48, opened 2016-01-26
Treat warnings as errors

8W9aG:
* Should only affect the open source CI


8W9aG:
@jgavris @rastersize @dflems @JensAyton

rastersize:
Hmm, won’t that affect all builds of the project? Or will our build-system “overwrite” it?

8W9aG:
Our build-system should overwrite it (since it will overwrite the project.xcconfig and point it to spotify_shared.xcconfig)

rastersize:
@8W9aG ok, :tada: :+1:

PR #49, opened 2016-01-26
Test Retry Authorisation

8W9aG:
* Tests the case of retrying after a single failure on authorisation errors from the server


8W9aG:
@jgavris @rastersize @dflems @JensAyton

jgavris:
:+1: Crushing that coverage

PR #56, opened 2016-01-29
Add testCancellingLoadsOnDealloc

8W9aG:
* Sanity checks that we don't crash when deallocating
while having requests in flight


8W9aG:
@rastersize @jgavris @dflems

rastersize:
I think this test will be _very_ difficult to implement without some major hackory. As with ARC doing `self.service = nil;` is not actually guaranteed to instantly release and dealloc the object. Instead that will happen eventually (the only guarantee).

I remember Anton trying to do a test like this internally. Eventually he ended up scrapping the idea.

rastersize:
After discussion outside of GitHub: :+1:

PR #58, opened 2016-01-29
Change iOS devices used to exercise more slices

rastersize:
- Trying to hit all of the arch slices – armv7, armv7s and arm64 – as well as 32 vs. 64-bit.

@8W9aG @jgavris @dflems


rastersize:
Coverage “dropped” for some reason!?

8W9aG:
:+1:

PR #62, opened 2016-01-29
Add SPTDataLoaderServiceTestNoARC

8W9aG:
* Allows us to test dealloc in a more controlled way
* I believe assuming ARC will deallocate at a specific
point is causing the CI unpredictability


8W9aG:
@rastersize @dflems @jgavris

rastersize:
Still think this test is weird. As the only way the test can “fail” is by crashing the entire test suite. Which is not amazing. But better than the previous ARC test. :+1: if you feel we need it.

8W9aG:
Going to merge this despite coverage going down, it went down due to the dealloc test being too flakey

PR #63, opened 2016-01-29
Delegate challenges to the default handling

8W9aG:
* We should not cancel authentication challenges
by default when we do not allow all certificates


8W9aG:
@rastersize @dflems @jgavris

rastersize:
:+1:

PR #65, opened 2016-01-30
Add test for different delegate queue

8W9aG:
* Keep that coverage up up up


8W9aG:
@rastersize @dflems @jgavris

jgavris:
:+1:

PR #66, opened 2016-01-30
Add code example for block based callbacks

8W9aG:
* Shows the reader how to get block based callbacks
for the SPTDataLoader delegates


8W9aG:
@rastersize @jgavris @dflems

jgavris:
:+1: But I'm pretty :-1: on this API. Kind of making a mockery of block API's.

8W9aG:
@jgavris can you explain?

jgavris:
If I wanted to use a block based API, I would prefer to just do `performRequest:completionBlock:`. You're telling me I have to implement a delegate protocol and actually call the block myself?

8W9aG:
Yeh, the aim is definitely to discourage the use of blocks when it comes to long network calls (for the reasons in the README)

jgavris:
If you want to discourage use because it's relatively easier to make mistakes by capturing references, why add this example though?

8W9aG:
Because the README makes a reference to the fact that sometimes block based callbacks are very useful depending on the circumstance (however it also states that those circumstances are rare). Having said that, if we are to acknowledge that this can be a good pattern when done correctly we should provide an example on how this can be done. I do believe that there should be some friction to using this pattern due to its drawbacks.

PR #70, opened 2016-01-31
Add Rate Limiting Specific Endpoints to README

8W9aG:
* Also adds a Readme Score coming from this site: http://scoreme.report/
* Adds the usage example with code in order to make it easy for the user of the library to search for this example in the README
* This usage example is not included in the demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 [![Cocoapods](https://img.shields.io/cocoapods/v/SPTDataLoader.svg)](https://cocoapods.org/?q=SPTDataLoader)
 [![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
 [![Spotify FOSS Slack](https://slackin.spotify.com/badge.svg)](https://slackin.spotify.com)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,6 +147,15 @@ didReceiveDataChunk:(NSData *)data
 ```
 Be sure to render YES in your delegate to tell the data loader that you support chunks, and to set the requests chunks property to YES.
 
+### Rate limiting specific endpoints
+If you specify a rate limiter in your service, you can give it a default requests per second metric which it applies to all requests coming out your app. (See "Creating the SPTDataLoaderService"). However, you can also specify rate limits for specific HTTP endpoints, which may be useful if you want to forcefully control the rate at which clients can make requests to a backend that does large amounts of work.
```
rastersize: How about linking references up?

```md
(See “[Creating the `SPTDataLoaderService`](#creating-the-sptdataloaderservice)”
```

That change also changes to use [correct quote characters](http://smartquotesforsmartpeople.com/).

rastersize:
:+1: although the suggested linking would be appreciated :blush:

8W9aG:
Code coverage once again a red herring (I believe there is a problem with the rand function of the exponential timer test)

PR #71, opened 2016-01-31
Add test for nil completion block in willCacheResponse

8W9aG:
* Ensure that we don't crash
* Also remove redundant isKindOfClass check


8W9aG:
@rastersize @dflems @jgavris

rastersize:
:+1:

PR #74, opened 2016-01-31
Add "Using the jittered exponential timer" to README

8W9aG:
* Not obvious how to use this API from looking at
the demo and the current README


8W9aG:
@rastersize @dflems @jgavris

jgavris:
:+1:

PR #75, opened 2016-02-01
Add testErrorDelegateCallbackWhenMismatchInChunkSupport

8W9aG:
* Test that an error is thrown when the data loader
attempts to load a request that requires chunks but
the delegate does not support chunks
* Removes previous assertion in favor of a less
noisy delegate error callback approach

### DIFF
--- a/SPTDataLoader/SPTDataLoader.m
+++ b/SPTDataLoader/SPTDataLoader.m
@@ -24,6 +24,7 @@
 #import "SPTDataLoaderResponse.h"
 #import "SPTDataLoaderRequestResponseHandler.h"
 #import "SPTDataLoaderDelegate.h"
+#import "SPTDataLoaderResponse+Private.h"
 
 @interface SPTDataLoader ()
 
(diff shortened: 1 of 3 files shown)


8W9aG:
@rastersize @dflems @jgavris

SPTDataLoader/SPTDataLoader.m
```diff
@@ -71,9 +72,18 @@ - (void)executeDelegateBlock:(dispatch_block_t)block
 {
     SPTDataLoaderRequest *copiedRequest = [request copy];
     id<SPTDataLoaderDelegate> delegate = self.delegate;
+
+    // Cancel the request immediately if it requires chunks and the delegate does not support that
     if ([delegate respondsToSelector:@selector(dataLoaderShouldSupportChunks:)] && copiedRequest.chunks) {
```
rastersize: Shouldn't you now also message the delegate in the if-statement to check if it supports chunks?

rastersize:
Could we add a test for the happy case?

SPTDataLoader/SPTDataLoader.m
8W9aG: Yes you are totally right

8W9aG:
@rastersize can do

8W9aG:
@rastersize happy case is tested in testRelayReceivedDataChunkToDelegate

rastersize:
@8W9aG Oh right, nice.

:+1:

PR #81, opened 2016-02-02
Add GIFs to high level sections in README

8W9aG:
* ScoreMe gives higher scores for READMEs with GIFs
so lets add couple in


8W9aG:
@rastersize @jgavris @dflems

iOSCowboy:
:dancer:

dflems:
this seems unnecessary. do we really care about an arbitrary readme score?

rastersize:
I agree with @dflems here, we've already reached 100/100 points. This here seems to mostly hurt readability and as such the quality of the readme. At least that's my opinion.

8W9aG:
Agreed

PR #82, opened 2016-02-02
Separate frameworks targets into new project

8W9aG:


rastersize:
Since we're recreating this project. Could we move all the different info.plist files into the same directory? Feels silly to have four directories with a single file in each. Better with one directory which contains all four of the files.

8W9aG:
@rastersize Do you mean rename all the Info.plist files to Info-Framework-IOS.plist or something?

rastersize:
@8W9aG If we can share the file, like now, that’s even nicer.

I think we should remove the local override of `INSTALL_PATH` and `DYLIB_INSTALL_NAME_BASE` from the Xcode project file. Instead `INSTALL_PATH` should be set to `@rpath` in the xcconfig.

Same with `LD_RUNPATH_SEARCH_PATHS` which should have its value from the xcconfig. The value should be `$(inherited) @executable_path/Frameworks @loader_path/Frameworks`

Remove the per-target override of `CLANG_MODULES_AUTOLINK` as it’s set in the xcconfig file.

rastersize:
:+1:

PR #85, opened 2016-02-03
Fix an intermittent warning in Xcode 7.3 beta

JensAyton:
This slightly changes behaviour: the default is used only if there is no override, not if the override is zero. The previous behaviour was undocumented an probably surprising, although I’m not sure what you’d expect to happen if you set zero.

@8W9aG @rastersize @wahni


8W9aG:
:+1:

tonylazarew:
:+1:

PR #89, opened 2016-02-05
Add langauge standards to xcconfig

rastersize:
To match what we have internally.

@8W9aG @dflems @jgavris @JensAyton


jgavris:
:+1:

PR #90, opened 2016-02-07
Add nullability & generic type annotations to all classes

JohnSundell:
This PR adds modern Objective-C annotations to all SPTDataLoader classes, both private and public.

Changes include:
- Adding `NS_ASSUME_NONNULL_BEGIN/END` compiler instructions to all files.
- Where objects are allowed to be `nil`, add the `nullable` annotation.
- Missing `nil`-checks and corresponding `_Nonnull` casts added.
- Some missing assignments to required variables added in tests.
- Where changes required to fully conform to the nullability annotations were considered too big for this commit, or where whether passing `nil` would cause a crash was tested, `#pragma` warning silencers were added.
- Generic annotations were added to all collections, both public & private.

Note: Whether a parameter should be `nullable` or not was determined by inspecting the surrounding code and the tests.

This change heavily improves SPTDataLoader’s usability in Swift, since arguments that before were typed as `[AnyObject]` now get full typing. Nullable properties and arguments are now also correctly treated as optionals in Swift.

This change also has the added benefit of adding more strong typing and safety to the implementation.

### DIFF
--- a/SPTDataLoaderTests/SPTDataLoaderFactoryTest.m
+++ b/SPTDataLoaderTests/SPTDataLoaderFactoryTest.m
@@ -113,7 +113,7 @@
     SPTDataLoaderRequest *request = [SPTDataLoaderRequest new];
     [self.factory requestResponseHandler:requestResponseHandler performRequest:request];
     SPTDataLoaderResponse *response = [SPTDataLoaderResponse dataLoaderResponseWithRequest:request response:nil];
-    [self.factory receivedDataChunk:nil forResponse:response];
+    [self.factory receivedDataChunk:[NSData new] forResponse:response];
     XCTAssertEqual(requestResponseHandler.numberOfReceivedDataRequestCalls, 1u, @"The factory did not relay a received data chunk response to the correct handler");
 }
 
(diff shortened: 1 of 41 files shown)


JohnSundell:
@8W9aG @jgavris @rastersize

SPTDataLoaderTests/SPTDataLoaderFactoryTest.m
```diff
@@ -165,21 +165,27 @@ - (void)testOfflineChangesCachePolicy
 - (void)testRelayToDelegateWhenPerformingRequest
 {
     SPTDataLoaderRequest *request = [SPTDataLoaderRequest new];
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wnonnull"
     [self.factory requestResponseHandler:nil performRequest:request];
+#pragma clang diagnostic pop
```
jgavris: I like `_Pragma("clang diagnostic push")`, it maintains the indentation level.

jgavris:
:+1:

rastersize:
Big like :+1:

SPTDataLoaderTests/SPTDataLoaderFactoryTest.m
JohnSundell: I like it too. In this case though I kind of want the warning silencers to be an eyesore, because we should ideally fix the issue and remove them. I just added them now since it was the smaller change, and I wanted to increase my confidence in the commit my not altering the tests too much. Feel free to change this to whatever code style you prefer though, for me it doesn't matter.

PR #91, opened 2016-02-08
Move build settings to xcconfig and normalize it

rastersize:
- Normalize the formatting in the `spotify_os.xcconfig`.
- Enables clang modules from the config.
- Moves almost all, if not all, shared settings to xcconfig.
- Fixes `ENABLE_NS_ASSERTIONS` not being set in the config.
- Fixes certain warnings found when building using Xcode 7.3b2 (so I could verify that this works).
- **Bumps OS X deployment target to 10.9.**
  - We’re using API which was introduced in 10.9. Since that version is so old by now I don’t see any reason to fix that usage and instead just bump the deployment target.

@8W9aG @JohnSundell @jgavris @dflems


jgavris:
:+1:

spotify_os.xcconfig
```diff
@@ -5,23 +5,52 @@ VALID_ARCHS = i386 x86_64 armv7s armv7 arm64 armv7k
...
+MACOSX_DEPLOYMENT_TARGET = 10.9
+
 ENABLE_NS_ASSERTIONS_Debug = YES
 ENABLE_NS_ASSERTIONS_Release = NO
+ENABLE_NS_ASSERTIONS = $(ENABLE_NS_ASSERTIONS_$(CONFIGURATION))
+
 ONLY_ACTIVE_ARCH_Debug = YES
 ONLY_ACTIVE_ARCH_Release = NO
 ONLY_ACTIVE_ARCH = $(ONLY_ACTIVE_ARCH_$(CONFIGURATION))
+DEBUG_INFORMATION_FORMAT_Debug = dwarf
```
jgavris: :+1:

PR #92, opened 2016-02-08
Remove NSObject type info for userInfo

rastersize:
It was causing issues with certain usage patterns where a non-NSObject subclass was saved in the userInfo dictionary. For example the [block interface hacking](https://github.com/spotify/SPTDataLoader#why-no-block-interface).

:crying_cat_face: 

@JohnSundell


JohnSundell:
:cry: :+1: :bow:

PR #93, opened 2016-02-08
Tweak the CI scripts a bit

rastersize:
- Extract project and scheme values for Slather from the scripts to the config file.
- Extract podspec name from the scripts to the config file.
- Make it easier to run CI script locally by setting the TRAVIS environment variable to a default (only if not yet set).

@jgavris @dflems


jgavris:
:+1:

PR #94, opened 2016-02-10
Support projects that don't enable clang modules

dflems:
Projects that don't have clang modules enabled currently can't support including our project. Switching to `#import` will work with both cases. When modules are enabled, these are implicitly converted to clang module imports.

@8W9aG @rastersize @jgavris


adamhowardprice:
:+1:

jgavris:
:+1:

Issue #104, opened 2016-02-19
@synthesize of 'weak' property is only allowed in ARC or GC mode


loretoparisi:
```
/MusiXmatch/3rdParty/SPTDataLoader/SPTDataLoader/SPTDataLoaderCancellationTokenImplementation.m:57:13: @synthesize of 'weak' property is only allowed in ARC or GC mode
```

Is the library a ARC project?

rastersize:
Hey @loretoparisi,

It’s indeed an ARC project. It looks like you need to init our submodules, which contains an Xcode config file defining almost all of our settings. Including the “use ARC” one.

Running the following command from the root of the project and then re-opening the Xcode workspace should resolve the issue:
```shell
git submodule update --init
```

loretoparisi:
@rastersize right I initialized the submodules, cleaned up, now I get

```
/MusiXmatch/3rdParty/SPTDataLoader/demo/PlaylistsViewModel.m:113:76: error: implicit conversion from nullable pointer 'NSData * _Nullable' to non-nullable pointer type 'NSData * _Nonnull' [-Werror,-Wnullable-to-nonnull-conversion]
    NSDictionary *jsonDictionary = [NSJSONSerialization JSONObjectWithData:response.body
```

loretoparisi:
Should'n it an unwrap like

```Objective-C
NSDictionary *jsonDictionary=nil;
    NSData *responseBody=response.body;
    if(responseBody) {
        jsonDictionary = [NSJSONSerialization JSONObjectWithData:responseBody
                                                                   options:NSJSONReadingAllowFragments
                                                                     error:&error];
    }
    
    if (!jsonDictionary) {
        NSLog(@"Error: %@", error);
        return;
    }
```

loretoparisi:
I was able to fix after unwrapping all those ones like in AppDelegate.m where I had two:

```Objective-c
NSError *error = nil;
    NSDictionary *oauthTokenDictionary = nil;
    NSData *responseBody=response.body;
    if(responseBody) {
       oauthTokenDictionary = [NSJSONSerialization JSONObjectWithData:responseBody
                                                                         options:NSJSONReadingAllowFragments
                                                                           error:&error];
    }
```

and in 

```Objective-C
- (void)dataLoader:(SPTDataLoader *)dataLoader didReceiveErrorResponse:(SPTDataLoaderResponse *)response
{
    @synchronized(self.pendingRequests) {
[...]

PR #105, opened 2016-02-19
Fix nullability issues in demo project

rastersize:
This PR fixes the nullability issues in the demo project pointed out by @loretoparisi in #104, thank you!

@8W9aG @dflems @adamhowardprice @loretoparisi

### DIFF
--- a/SPTDataLoaderDemo.xcodeproj/project.pbxproj
+++ b/SPTDataLoaderDemo.xcodeproj/project.pbxproj
@@ -317,6 +317,7 @@
 					"$(inherited)",
 					include,
 				);
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				MTL_ENABLE_DEBUG_INFO = YES;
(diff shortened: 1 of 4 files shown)


SPTDataLoaderDemo.xcodeproj/project.pbxproj
```diff
@@ -317,6 +317,7 @@
 					"$(inherited)",
 					include,
 				);
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
```
8W9aG: Hmmm shouldn't this be 7.0?

rastersize: No, as the demo project is using iOS 8.0 and up API. Will not build with Xcode 7.3.

8W9aG: Ah true!

PR #106, opened 2016-02-22
Fix nil-returns from methods with nonnull return type

rastersize:
Caught by running the analyzer bundled with Xcode 7.3 beta 3. Most of the problems were due to the early-return segment of the init methods:
early return nil analyzer warning

Also fixed a the type of some methods that should return `nil` as well as updated the documentation for those.

@spotify/objc-dev

### DIFF
--- a/SPTDataLoader/SPTDataLoader.m
+++ b/SPTDataLoader/SPTDataLoader.m
@@ -30,8 +30,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface SPTDataLoader ()
 
-@property (nonatomic, strong) NSHashTable<id<SPTDataLoaderCancellationToken>> *cancellationTokens;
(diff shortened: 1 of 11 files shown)


8W9aG:
:+1:

iOSCowboy:
I personally support this change, but... shouldn't we update our Obj-c guidelines first to be consistent?

rastersize:
@iOSCowboy: We can hold off merging until we’ve officially updated them :smiley_cat: 

Just bothered me that I couldn’t build the project without getting all these warnings :stuck_out_tongue_closed_eyes:

iOSCowboy:
@rastersize: Totally. It's like walking with a pebble in your shoe.

rastersize:
Style in our guidelines has been updated, spotify/ios-style#5. Merging this change.

iOSCowboy:
You never lifted the finger from the trigger for this one @rastersize

PR #107, opened 2016-02-29
Move -debugDescription implementation to -description method for improved debugability

rastersize:
This PR moves the `-debugDescription` implementation to the `-description` method instead. The reasoning being that it makes debugging easier if a request or response object is logged instead of printed using the debugger.

Also aligns the format with Apple’s standard (`<ClassName: 0xDEADBEEF property-1 = "string value"; property-2 = 42>`).

@spotify/objc-dev

### DIFF
--- a/SPTDataLoader/SPTDataLoaderRequest.m
+++ b/SPTDataLoader/SPTDataLoaderRequest.m
@@ -177,9 +177,9 @@ static NSString * NSStringFromSPTDataLoaderRequestMethod(SPTDataLoaderRequestMet
     return [languageHeaderValues componentsJoinedByString:SPTDataLoaderRequestLanguageHeaderValuesJoiner];
 }
 
-- (NSString *)debugDescription
+- (NSString *)description
 {
-    return [NSString stringWithFormat:@"%@ { URL: %@ }", [super debugDescription], self.URL];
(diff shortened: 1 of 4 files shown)


8W9aG:
:+1:

iOSCowboy:
Why not both? If we log these objects we will use `-description`, but we might still want to `po` individual objects in the console. 

I would suggest returning the `-description` string from `-debugDescription`.

Note: I'm assuming the default implementation of `-debugDescription` just prints the object and its address.

rastersize:
@iOSCowboy: This will work with `po` in the debugger as well. Unless we want to output something extra, or special, when printed from the debugger all that’s needed it so implement `-description`. See the documentation for `-[NSObject debugDescription]`:

> `- debugDescription`
> Returns a string that describes the contents of the receiver for presentation in the debugger.
>
> `NSObject` implements this method by calling through to the `description` method. Thus, by default, an object’s debug description is the same as its description. However, you can override `debugDescription` if you want to decouple these.

– [https://developer.apple.com/…/NSObject/debugDescription](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Protocols/NSObject_Protocol/index.html#//apple_ref/occ/intfm/NSObject/debugDescription)

iOSCowboy:
Perfect! Thanks for the note, @rastersize.

PR #108, opened 2016-03-26
Podspec fixes

rastersize:
Fixes some issues, adds some more stuff and standardizes the podspec.

- Adds a link to our [Spotify Engineering Twitter](http://twitter.com/spotifyeng) account.
- Fixes iOS deployment target. It was set to 6.0 even though we only support 7.0 and up.
- Fixes use of incorrect apostrophe character.
- Re-formats the podspec making sure we use only one type of quote character. As well as some other small changes to the formatting.

@spotify/objc-dev


8W9aG:
:+1:

PR #113, opened 2016-04-22
getting redirect_uri from ClientKeys.h

mauriciomaniglia:
Little adjustment on the appDelegate from the demo.


8W9aG:
👍 Thanks Maurício

PR #116, opened 2016-05-03
Remove Slather

rastersize:
As we don’t use it anymore. Also bumps other gems.


rastersize:
@spotify/objc-dev

8W9aG:
👍

rastersize:
I’ll investigate why Codecov fails to report the status to GHE later. For now, merge time.

PR #118, opened 2016-06-23
Fix a potential issue where we could end up with duplicate symbols

rastersize:
Only happens with certain build system configurations. This PR resolves it by always using a very specific `HEADER_SEARCH_PATHS` that **MUST NOT** include `$(inherited)`.

@8W9aG @dflems @spotify/objc-dev


8W9aG:
👍

PR #119, opened 2016-06-27
Fix BAD_POINTER_COMPARISON (infer warning)

mumme:
Should fix this:

SPTDataLoaderRateLimiter.m 
`131`: Implicitly checking whether `NSNumber` pointer value is nil at line `131`, column `16`. Did you mean to compare against the unboxed value instead? Please either explicitly compare value to nil or use one of the `NSNumber` accessors before the comparison.


8W9aG:
👍

rastersize:
👍

PR #122, opened 2016-08-16
Fix cancellation race condition

8W9aG:
* Currently when the user cancels the cancellation token we receive this delegate callback in the SPTDataLoaderService
* The SPTDataLoaderService then proceeds to cancel the associated data task
* This would then asynchronously cancel the request, once cancelled that message was forwarded up to the SPTDataLoader which would then alert the delegate
* The problem occurs because of the asynchronousity of the data task cancellation, we could theoretically receive an error or success message before we cancel the data task despite the fact that the cancellation came first
* We will solve this problem by cancelling requests in SPTDataLoader instead and disallowing them  to bubbled up to the UI immediately once cancelled, and perform the actual cancel further down the stack

### DIFF
--- a/SPTDataLoader/SPTDataLoaderService.m
+++ b/SPTDataLoader/SPTDataLoaderService.m
@@ -200,6 +200,21 @@ requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseH
 }
 
 - (void)requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
+                 cancelRequest:(SPTDataLoaderRequest *)request
+{
+    NSArray *handlers = nil;
+    @synchronized(self.handlers) {
+        handlers = [self.handlers copy];
+    }
+    for (SPTDataLoaderRequestTaskHandler *handler in handlers) {
+        if ([handler.request isEqual:request]) {
+            [handler.task cancel];
+            break;
+        }
+    }
+}
+
+- (void)requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
              authorisedRequest:(SPTDataLoaderRequest *)request
 {
     [self performRequest:request requestResponseHandler:requestResponseHandler];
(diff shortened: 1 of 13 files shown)


8W9aG:
@nataliq

8W9aG:
Tests are coming, verify whether this architecture sounds sane first

SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -187,24 +184,34 @@ - (void)cancelAllLoads
...
+}
+
+- (void)requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
+                 cancelRequest:(SPTDataLoaderRequest *)request
+{
+    NSArray *handlers = nil;
+    @synchronized(self.handlers) {
+        handlers = [self.handlers copy];
+    }
+    for (SPTDataLoaderRequestTaskHandler *handler in handlers) {
```
nataliq: Shouldn't you use the provided `requestResponseHandler` instead?

SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -187,24 +184,34 @@ - (void)cancelAllLoads
 
 #pragma mark SPTDataLoaderRequestResponseHandlerDelegate
 
-- (nullable id<SPTDataLoaderCancellationToken>)requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
-                                                       performRequest:(SPTDataLoaderRequest *)request
+- (void)requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
```
nataliq: It makes more sense to combine `requestResponseHandler:performRequest:` and `performRequest:requestResponseHandler:`, this way it is confusing.

SPTDataLoader/SPTDataLoaderService.m
8W9aG: A request task handler isn't the same as a requestResponseHandler

SPTDataLoader/SPTDataLoaderService.m
8W9aG: The separation is needed due to how the authentication logic works

KeenanCassidy:
👍

PR #123, opened 2016-08-16
Do not increase the unique identifier barrier on copy

8W9aG:
* Prevents requests from increasing by 2 on each
perform request rather than just by 1


dflems:
This seems like a good use case for using `OSAtomic`, specifically `OSAtomicIncrement64`

8W9aG:
OSAtomicIncrement64 is not implemented in iOS 7.0, just iOS 7.1, so it's not possible to use it at the moment

KeenanCassidy:
👍

PR #124, opened 2016-10-06
Project file cleanup

JensAyton:
Make Xcode 8 happier.

@8W9aG


8W9aG:
👍

SPTDataLoader.xcodeproj/project.pbxproj
```diff
@@ -531,6 +531,7 @@
 				GCC_SYMBOLS_PRIVATE_EXTERN = NO;
 				HEADER_SEARCH_PATHS = "\"$(SRCROOT)/include\"";
 				MTL_ENABLE_DEBUG_INFO = YES;
+				ONLY_ACTIVE_ARCH = YES;
```
rastersize: I think we could remove this form the project file, since it’s [defined in our xcconfig](https://github.com/spotify/SPTDataLoader/blob/master/ci/spotify_os.xcconfig#L32-L34).

JensAyton: This is true.

PR #127, opened 2016-10-24
Nullability annotations for user agent, rate limiter and resolver

joetam:
Adding nullability annotations to reflect what it says in the README under `Creating the SPTDataLoaderService` so clients don't get a warning for passing in nil.

> Note that you can provide all these as nils if you are so inclined, it may be for the best to use nils until you identify a need for these different configuration options.


joetam:
@spotify/objc-dev

PR #130, opened 2016-11-29
Add @rpath to install name base

8W9aG:
* Allows the framework to be used from other Carthage frameworks


8W9aG:
@spotify/objc-dev @joetam

PR #131, opened 2016-11-29
Fix Retries

8W9aG:
* Handler was always being removed regardless of whether it was attempting to retry
* Not only that, but it attempted to retry on a task that had just completed
* Also fixes partial availability error on Xcode 8.1


8W9aG:
@spotify/objc-dev

KeenanCassidy:
+1

SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -308,14 +308,22 @@ - (void)URLSession:(NSURLSession *)session
 didCompleteWithError:(nullable NSError *)error
 {
     SPTDataLoaderRequestTaskHandler *handler = [self handlerForTask:task];
+    handler.task = [self.session dataTaskWithRequest:handler.request.urlRequest];
```
8W9aG: That's okej by this time the task has expired

PR #132, opened 2016-12-05
Fix Race Conditions associated with new tasks for retries

8W9aG:
* Issue 1: The task property was nonatomic while being accessed from different threads, this is a no no
* Issue 2: There are race conditions which can lead to handlers not being associated with a task, this is solved


8W9aG:
@spotify/objc-dev

adamhowardprice:
👍

PR #137, opened 2017-05-22
Add module map

JensAyton:


JensAyton:
@spotify/objc-dev

5at5ish:
👍

include/SPTDataLoader/module.modulemap
```diff
@@ -0,0 +1,3 @@
+module SPTDataLoader {
+    umbrella "."
```
rastersize: Could we use the [`SPTDataLoaderConvenience.h`](https://github.com/spotify/SPTDataLoader/blob/master/include/SPTDataLoader/SPTDataLoaderConvenience.h) umbrella header instead?

dflems: It's a breaking change, but we should probably rename `SPTDataLoaderConvenience.h` to `SPTDataLoader.h` and rename the existing `SPTDataLoader.h` to... something else.

PR #139, opened 2017-06-08
Move certificate resources into bundles

dflems:
This helps a bit with some build system stuff I'm working on. The folder references were causing some problems with these test resources, so I moved them into sub-bundles.

@rastersize @8W9aG


rastersize:
👍

PR #140, opened 2017-08-11
Use half open interval [0, 1) as original algorithm does

bitomaxsp:
Also such change avoids devision by zero that can happen in rare cases


8W9aG:
👍

bitomaxsp:
@spotify/objc-dev Why did build fail ? Can that be fixed? :)

dflems:
@8W9aG Looks like the cert bundled in the repo has expired... I think this is safe to merge. The tests that are failing are not related to the change.

PR #144, opened 2017-09-20
Revert "Add module map"

8W9aG:
Reverts spotify/SPTDataLoader#137

* I didn't check the Travis build output properly


8W9aG:
@spotify/objc-dev

cerihughes:
@rastersize @8W9aG - can you remember why we had to revert this? I need to use `SPTDataLoader` from a Swift project so was wondering what the process is now. Thx!

dflems:
@cerihughes I *think* we should be able to re-add this safely now.

PR #145, opened 2017-09-21
Add HEAD request method support

8W9aG:
* Adds ability for data loader to support HEAD requests
* Really useful for checking whether an item is available on a CDN without downloading the whole thing


8W9aG:
@spotify/objc-dev

PR #148, opened 2017-10-31
Delegates for reporting sent/received bytes

podkovyrin:


SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
```diff
@@ -239,11 +271,86 @@ - (void)completeIfInFlight
...
+                [requestResponseHandler updatedCountOfBytesReceived:self.task.countOfBytesReceived
+                                      countOfBytesExpectedToReceive:self.task.countOfBytesExpectedToReceive
+                                                         forRequest:self.request];
+            } else if ([keyPath isEqualToString:kCountOfBytesSent] || [keyPath isEqualToString:kCountOfBytesExpectedToSend]) {
+                [requestResponseHandler updatedCountOfBytesSent:self.task.countOfBytesSent
+                                     countOfBytesExpectedToSend:self.task.countOfBytesExpectedToSend
+                                                     forRequest:self.request];
+            }
+        }
+    }
```
erikprice: Should we call `[super observeValueForKeyPath:…]` in the `else` case?

podkovyrin: Calling super implementation is by the book but `SPTDataLoaderRequestTaskHandler` is the first class in the inheritance chain. So if anyone inherits from `SPTDataLoaderRequestTaskHandler` and makes its own default KVO implementation it will lead to the crash because calling `observeValueForKeyPath:ofObject:change:context:` on `NSObject` throws an `NSInternalInconsistency` exception.

PR #152, opened 2018-04-21
Add background download task support

SPTAndrewChang:
Rather than always using a data task to make requests, this change allows a "background policy" to be set on requests that determine when/if a background task is created:

- **Default**: Allow the system to fail in-flight requests in the background
- **OnDemand**: Upgrade data tasks to background tasks after receiving response headers
- **Always**: Use a background task at the start, don't return response headers/status code

Background tasks are important for long-running requests which may be killed by the operating system when the application loses foreground focus. It's possible to improve the reliability of such requests by using a download task instead.

This change is backwards compatible. The default policy (.Default) is to always use a data task as before.

### DIFF
--- a/SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
+++ b/SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
@@ -108,7 +108,11 @@ static NSUInteger const SPTDataLoaderRequestTaskHandlerMaxRedirects = 10;
         if (self.request.chunks) {
             [self.requestResponseHandler receivedDataChunk:dataRange forResponse:self.response];
         } else {
-            [self.receivedData appendData:dataRange];
+            if (!self.receivedData) {
+                self.receivedData = [dataRange mutableCopy];
+            } else {
+                [self.receivedData appendData:dataRange];
+            }
         }
     }];
 }
@@ -162,7 +166,7 @@ static NSUInteger const SPTDataLoaderRequestTaskHandlerMaxRedirects = 10;
 {
     self.response = [SPTDataLoaderResponse dataLoaderResponseWithRequest:self.request response:response];
     [self.requestResponseHandler receivedInitialResponse:self.response];
-    
+
     if ([response isKindOfClass:[NSHTTPURLResponse class]]) {
         NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
         if (httpResponse.expectedContentLength > 0) {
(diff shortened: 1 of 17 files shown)


SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
```diff
@@ -108,7 +108,11 @@ - (void)receiveData:(NSData *)data
         if (self.request.chunks) {
             [self.requestResponseHandler receivedDataChunk:dataRange forResponse:self.response];
         } else {
-            [self.receivedData appendData:dataRange];
+            if (!self.receivedData) {
+                self.receivedData = [dataRange mutableCopy];
```
erikprice: Does this change may fix a latent bug where we assumed that `receivedData` was always non-nil? Do we have a test that exposes this bug?

SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
```diff
@@ -162,9 +166,10 @@ - (NSURLSessionResponseDisposition)receiveResponse:(NSURLResponse *)response
 {
     self.response = [SPTDataLoaderResponse dataLoaderResponseWithRequest:self.request response:response];
     [self.requestResponseHandler receivedInitialResponse:self.response];
-    
+
+    NSHTTPURLResponse *httpResponse = nil;
```
erikprice: It's not clear why this local was moved.

SPTDataLoaderTests/SPTDataLoaderRequestTaskHandlerTest.m
```diff
@@ -204,4 +204,20 @@ - (void)testCancelledWhenReceivingCancelError
     XCTAssertTrue(self.handler.cancelled);
 }
 
+- (void)testUpgradeToDownloadTaskAferResponse
```
erikprice: `s/Afer/After/`

SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
SPTAndrewChang: It only applies to `backgroundPolicy = .Always` which doesn’t receive an initial response (and thus wasn’t initializing `receivedData`). I added a test for this behavior.

SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
SPTAndrewChang: There was a check for response headers but that change was (only partially) reverted. Should be fixed now.

SPTDataLoaderTests/SPTDataLoaderRequestTaskHandlerTest.m
SPTAndrewChang: Good catch, fixed.

8W9aG:
👍

PR #154, opened 2018-05-16
Report downloaded size using Content-Length

gitrema:
Use the Content-Length field of the HTTP response header first.
If missing fallback to the number of bytes reported by the task.
This is useful when compressed data is returned by the server.
In that case, the task will report the uncompressed data instead
of the compressed one that was actually downloaded.


gitrema:
@8W9aG

gitrema:
@8W9aG it seems I'm not authorized to merge the change. Can you do it for me?
Thanks.

PR #157, opened 2018-05-25
Fix broken tests

dflems:
This PR fixes the tests that were broken in #154.

@gitrema @8W9aG Can you please write some tests for the new logic?


dflems:
@rastersize

PR #165, opened 2019-01-28
Refactors removing cancellation tokens avoiding mutation of array during iterating it

valeriyvan:
While refactored code is correct, mutation of collection while iterating it always raises a warning.
Proposed change is more CPU effective as removal of multiple array elements in one call avoids
multiple buffer copying and possible array storage deallocation.
Price for that effectives - need for storage of indices in NSMutableIndexSet.


rastersize:
I had a look at the Travis-CI failure and it’s a false-positive. Going to merge anyhow (the same test succeeded with a different simulator in the same run).

PR #166, opened 2019-02-26
Add modulemap

dflems:


include/SPTDataLoader/module.modulemap
```diff
@@ -0,0 +1,19 @@
+module SPTDataLoader {
+    umbrella header "SPTDataLoader.h"
+
+    header "SPTDataLoaderAuthoriser.h"
```
JensAyton: Surely listing these as well as the `umbrella header` clause is redundant?

dflems: I dunno! I'm going by some examples I've seen elsewhere.

dflems: Figured it would be best to be as complete as possible.

JensAyton: I really don’t think so, this is effectively specifying headers twice which is at best unnecessary complexity

PR #172, opened 2019-08-12
Refuse new requests if session has become invalid

fabriziodemaria:
We have observed an issue where the underlying `NSURLSession` gets invalidated during App's shutdown and after that new attempts to create tasks for that session are initiated, causing a crash of the App with message: `Task created in a session that has been invalidated`.

This PR attempts to make the `SPTDataLoader` more forgiving in this scenario, rejecting new requests/task without crashing the App.

_Note: The solution currently proposed proved to be racy when a new task is created right after the `NSURLSession` is invalidated. Hence this PR is marked as Work-In-Progress while looking at other possible ways to fix the problem._


fabriziodemaria:
Stale PR, we can reopen if needed in the future.

PR #175, opened 2019-09-04
Add per-request waitsForConnectivity flag

JensAyton:
* Adds `waitForConnectivity` flag to `SPTDataLoaderRequest`
* Adds a mechanism to support multiple `NSURLSession`s and select between them for each request
* Routes `waitingForConnectivity` notification through the various layers

Missing:
- ~[ ] Tests~ — not effectively testable
- [x] Documentation
- [x] The request sent to the delegate is currently the internal copy of the request made by SPTDataLoader for reasons, this is bad.

### DIFF
--- a/Sources/SPTDataLoaderFactory.m
+++ b/Sources/SPTDataLoaderFactory.m
@@ -31,7 +31,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface SPTDataLoaderFactory () <SPTDataLoaderRequestResponseHandlerDelegate, SPTDataLoaderAuthoriserDelegate>
+@interface SPTDataLoaderFactory () <SPTDataLoaderRequestResponseHandlerDelegate, SPTDataLoaderAuthoriserDelegate, SPTDataLoaderRequestResponseHandler>
 
 @property (nonatomic, strong) NSMapTable<SPTDataLoaderRequest *, id<SPTDataLoaderRequestResponseHandler>> *requestToRequestResponseHandler;
 @property (nonatomic, strong, readwrite) dispatch_queue_t requestTimeoutQueue;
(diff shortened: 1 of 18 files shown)


JensAyton:
“The request sent to the delegate is currently the internal copy of the request made by SPTDataLoader for reasons, this is bad.”

Apparently this happens with all delegate methods and you have to use numerical identifiers if you want to tell requests apart. ¯\_(ツ)_/¯

Sources/SPTDataLoaderFactory.m
```diff
@@ -31,7 +31,7 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface SPTDataLoaderFactory () <SPTDataLoaderRequestResponseHandlerDelegate, SPTDataLoaderAuthoriserDelegate>
+@interface SPTDataLoaderFactory () <SPTDataLoaderRequestResponseHandlerDelegate, SPTDataLoaderAuthoriserDelegate, SPTDataLoaderRequestResponseHandler>
```
JensAyton: Bug fix: `SPTDataLoaderFactory` is used as a `SPTDataLoaderRequestResponseHandler` but didn’t formally conform to it.

Sources/SPTDataLoaderService.m
```diff
@@ -52,6 +54,10 @@ @interface SPTDataLoaderService () <SPTDataLoaderRequestResponseHandlerDelegate,
 @end
 
 @implementation SPTDataLoaderService
+{
+    NSURLSession *_nonWaitingSession;
+    NSURLSession *_waitingSession;
```
JensAyton: The `sessionForRequest:` mechanism could be extended to handle other session configuration flags, such as [`allowsCellularAccess`](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1409406-allowscellularaccess?language=objc), [`allowsExpensiveNetworkAccess`](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/3235752-allowsexpensivenetworkaccess?language=objc), and [`allowsConstrainedNetworkAccess`](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/3235751-allowsconstrainednetworkaccess?language=objc) (iOS 13 Low Data Mode).

If we do this, I’d suggest merging the flags into a bitfield, masking out the ones that aren’t available on the current OS, and using that as a key into a dictionary of `NSURLSession`s.

Sources/SPTDataLoaderRequestTaskHandler.h
```diff
@@ -73,6 +73,10 @@ NS_ASSUME_NONNULL_BEGIN
  @return The response object unless the request was cancelled or will be re-tried in which case `nil` is returned.
  */
 - (nullable SPTDataLoaderResponse *)completeWithError:(nullable NSError *)error;
+/**
+ Called when a request with waitsForConnectivity enters the waiting state
+*/
+- (void)noteWaitingForConnectivity;
```
JensAyton: This isn’t a very good name, but because `SPTDataLoaderRequestTaskHandler` doesn’t follow the delegate naming convention it’s hard to come up with a good verb form.

PR #178, opened 2019-09-13
Convert to github workflow

dflems:
Will update branch protection rules & webhooks once I have a +1


ci/run.sh
```diff
@@ -105,7 +106,22 @@ xcb "Run tests for tvOS" test \
 #
 # CODECOV
 #
+
+# output a bunch of stuff that codecov might recognize
```
rastersize: Is this because Codecov doesn’t support/recognize GitHub workflows yet?

dflems: The bash script doesn't recognize GH workflows as a CI provider. They have a GH action, but it's not as parameterizable as the bash script. Thinking about kicking a PR [upstream](https://github.com/codecov/codecov-bash) to add support for it.

PR #179, opened 2019-09-17
Add a property to indicate whether request is serviced from cache

dmakarov:


Sources/SPTDataLoaderService.m
```diff
@@ -237,6 +237,8 @@ - (void)performRequest:(SPTDataLoaderRequest *)request
     if (request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
         task = [self.session downloadTaskWithRequest:urlRequest];
     } else {
+        NSCachedURLResponse *cachedResponse = [self.session.configuration.URLCache cachedResponseForRequest:urlRequest];
+        request.isCached = cachedResponse != nil;
```
dflems: Is there no other way to know that a response came from the cache? It feels a little weird to mutate the _request_ for this, no? Wouldn't it make more sense as a property on the response object? `cameFromCache` - something like that?

dmakarov: I didn’t find a better way to find out if a response may be coming from the cache. It seems at this point in code the response isn’t available yet, that’s why the request is mutated. Moreover the request is used in request accounting reporting therefore it’s convenient to store the property in the request object. We need this property to avoid overreporting network traffic in request accounting messages.

dflems: Have you verified that the right value comes back in the delegate callbacks? IIRC the request is `copy`ed at some point.

dflems: We should probably use the async `getCachedResponseForDataTask:completionHandler:` method, and if we get a hit, we can short-circuit even pushing the data through the URL session and just fire the delegate callbacks manually. Maybe easier said than done.

dmakarov: Do you have examples of using `getCachedResponseForDataTask`? I can't seem to find any.

Sources/SPTDataLoaderRequestTaskHandler.m
```diff
@@ -198,7 +203,16 @@ - (BOOL)mayRedirect
...
+    [self.cache getCachedResponseForDataTask:(NSURLSessionDataTask*)self.task
+                           completionHandler:^(NSCachedURLResponse *cachedResponse) {
+        if (cachedResponse) {
+            [self.task cancel];
+            self.cancelled = YES;
+            [self receiveResponse:(NSURLResponse*)cachedResponse];
+        } else {
+            self.executionBlock();
+        }
+    }];
```
dmakarov: @dflems This is incomplete, but is this the direction I should explore?

Sources/SPTDataLoaderRequestTaskHandler.m
```diff
@@ -198,7 +209,23 @@ - (BOOL)mayRedirect
 - (void)start
 {
     self.started = YES;
-    self.executionBlock();
+    [self.session.configuration.URLCache getCachedResponseForDataTask:(NSURLSessionDataTask*)self.task
+                                                    completionHandler:^(NSCachedURLResponse *cachedResponse) {
+        NSLog(@"getCachedResponseForDataTask completionHandler got response %p", (__bridge void*)cachedResponse);
+        if (cachedResponse) {
+            [self.task cancel];
```
gitrema: this is dangerous. There is no way to check the etag is still valid without contacting the server, for example

gitrema: From the documentation is not clear if expired content in the cache will be served or discarded

dmakarov:
Closing this for now to rework with an approach that adds a special header when a response is stored in the cache, which we can later use to recognizes the responses received from the cache.

PR #180, opened 2019-10-04
Add a method to check whether a response is returned from cache

dmakarov:


Sources/SPTDataLoaderURLCache.m
```diff
@@ -0,0 +1,60 @@
...
+@implementation SPTDataLoaderURLCache
+
+#pragma mark SPTDataLoaderURLCache
+
+- (NSCachedURLResponse *)addCachedHeaderToResponse:(NSCachedURLResponse *)cachedResponse
+{
+    if ([cachedResponse.response isKindOfClass:[NSHTTPURLResponse class]]) {
+        NSHTTPURLResponse *response = (NSHTTPURLResponse*) cachedResponse.response;
+        NSMutableDictionary *headers = [response.allHeaderFields mutableCopy];
+        headers[@"Cached"] = @"YES";
```
gitrema: For the header name I would use a more weird string to avoid clashing with an existent one. Something like @"X-SPTDataLoader-Cache-Header" and put it as a constant in a shared header file.

gitrema:
It seems some tests can be added

Sources/SPTDataLoaderURLCache.m
dmakarov: updated.

Sources/SPTDataLoaderResponse.m
```diff
@@ -62,8 +63,19 @@ - (instancetype)initWithRequest:(SPTDataLoaderRequest *)request response:(nullab
                                              code:httpResponse.statusCode
                                          userInfo:nil];
             }
-            _responseHeaders = httpResponse.allHeaderFields;
+            NSString* value = [_responseHeaders valueForKey:SPTDataLoaderURLCacheSpecialHeader];
```
gitrema: Did you mean to use `httpResponse.allHeaderFields` instead of `_responseHeaders` here?
Otherwise it looks to me you are using an uninitialised member

dmakarov: good catch. thank you. working on tests now.

dmakarov:
@gitrema Added tests.

Sources/SPTDataLoaderURLCache.h
```diff
@@ -0,0 +1,31 @@
...
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ */
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+static NSString * const SPTDataLoaderURLCacheSpecialHeader = @"X-SPTDataLoader-Cache-Header";
```
gitrema: I think it would be safer to put the definition in the .m file and just declare the symbol as external

Sources/SPTDataLoaderResponse.m
```diff
@@ -63,6 +65,13 @@ - (instancetype)initWithRequest:(SPTDataLoaderRequest *)request response:(nullab
                                          userInfo:nil];
             }
             _responseHeaders = httpResponse.allHeaderFields;
+            NSString* value = [_responseHeaders valueForKey:SPTDataLoaderURLCacheSpecialHeader];
+            if (value && [value isEqualToString:@"YES"]) {
+                NSMutableDictionary *headers = [httpResponse.allHeaderFields mutableCopy];
+                [headers removeObjectForKey:@"Cached"];
```
gitrema: Shouldn't be that one SPTDataLoaderURLCacheSpecialHeader ?

Sources/SPTDataLoaderURLCache.h
dmakarov: good point

Sources/SPTDataLoaderResponse.m
dmakarov: oops. indeed. good catch

PR #181, opened 2019-10-09
Turn SPTDataLoaderResponseHTTPStatusCode back to a regular enum

JensAyton:
In #174 I changed this from `NS_ENUM` to `NS_ERROR_ENUM` for modern style and better bridging to Swift. However, it turns out that `SPTDataLoaderResponseHTTPStatusCode` is used _both_ as an `NSError` code and in other contexts, which is terribad and breaks things.

See https://github.com/spotify/SPTDataLoader/commit/5f62f06b955dac2cd5ccb4dea27be73dad5d39af#r35424354

@dmakarov @gitrema

### DIFF
--- a/include/SPTDataLoader/SPTDataLoaderResponse.h
+++ b/include/SPTDataLoader/SPTDataLoaderResponse.h
@@ -22,10 +22,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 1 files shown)


dmakarov:
can someone merge this please, @JensAyton @gitrema? thanks

PR #186, opened 2019-12-22
Allow SPTDataLoader clients to pass callbacks to handle request respo…

dmakarov:
…nses

### DIFF
--- a/Sources/SPTDataLoader.m
+++ b/Sources/SPTDataLoader.m
@@ -37,6 +37,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, readonly) NSMutableArray<id<SPTDataLoaderCancellationToken>> *cancellationTokens;
 @property (nonatomic, strong, readonly) NSMutableArray<SPTDataLoaderRequest *> *requests;
 @property (nonatomic, strong, readonly) id<SPTDataLoaderCancellationTokenFactory> cancellationTokenFactory;
+@property (nonatomic, strong) NSMapTable *requestHandlers;
 
 @end
 
@@ -62,6 +63,7 @@ NS_ASSUME_NONNULL_BEGIN
         _cancellationTokens = [NSMutableArray new];
         _delegateQueue = dispatch_get_main_queue();
         _requests = [NSMutableArray new];
+        _requestHandlers = [NSMapTable strongToStrongObjectsMapTable];
     }
     return self;
 }
@@ -130,6 +132,26 @@ NS_ASSUME_NONNULL_BEGIN
     return cancellationToken;
 }
 
+- (nullable id<SPTDataLoaderCancellationToken>)performRequest:(SPTDataLoaderRequest *)request
+                                               withCompletion:(void (^)(SPTDataLoaderResponse *response))completion
+{
+    SPTDataLoaderRequest *copiedRequest = [request copy];
+    id<SPTDataLoaderCancellationToken> cancellationToken = [self.cancellationTokenFactory createCancellationTokenWithDelegate:self
+                                                                                                                 cancelObject:copiedRequest];
+    copiedRequest.cancellationToken = cancellationToken;
+    @synchronized(self.cancellationTokens) {
+        [self.cancellationTokens addObject:cancellationToken];
+    }
+    @synchronized(self.requests) {
+        [self.requests addObject:copiedRequest];
+    }
+    @synchronized(self.requestHandlers) {
+        [self.requestHandlers setObject:completion forKey:request];
+    }
+    [self.requestResponseHandlerDelegate requestResponseHandler:self performRequest:copiedRequest];
+    return cancellationToken;
+}
+
 - (void)cancelAllLoads
 {
     NSArray *cancellationTokens = nil;
@@ -175,11 +197,16 @@ NS_ASSUME_NONNULL_BEGIN
     if (![self isRequestExpected:response.request]) {
         return;
     }
-    
-    [self executeDelegateBlock: ^{
+    [self executeDelegateBlock:^{
         [self.delegate dataLoader:self didReceiveSuccessfulResponse:response];
+        void (^handler)(SPTDataLoaderResponse *) = [self.requestHandlers objectForKey:response.request];
+        if (handler) {
+            handler(response);
+            @synchronized(self.requestHandlers) {
+                [self.requestHandlers removeObjectForKey:response.request];
+            }
+        }
     }];
-
     [self removeRequest:response.request];
 }
 
@@ -188,11 +215,16 @@ NS_ASSUME_NONNULL_BEGIN
     if (![self isRequestExpected:response.request]) {
         return;
     }
-
-    [self executeDelegateBlock: ^{
+    [self executeDelegateBlock:^{
         [self.delegate dataLoader:self didReceiveErrorResponse:response];
+        void (^handler)(SPTDataLoaderResponse *) = [self.requestHandlers objectForKey:response.request];
+        if (handler) {
+            handler(response);
+            @synchronized(self.requestHandlers) {
+                [self.requestHandlers removeObjectForKey:response.request];
+            }
+        }
     }];
-
     [self removeRequest:response.request];
 }
 
@@ -201,13 +233,20 @@ NS_ASSUME_NONNULL_BEGIN
     if (![self isRequestExpected:request]) {
         return;
     }
-
     if ([self.delegate respondsToSelector:@selector(dataLoader:didCancelRequest:)]) {
-        [self executeDelegateBlock: ^{
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didCancelRequest:request];
         }];
     }
-
+    [self executeDelegateBlock:^{
+        void (^handler)(SPTDataLoaderResponse *) = [self.requestHandlers objectForKey:request];
+        if (handler) {
+            handler(nil);
+            @synchronized(self.requestHandlers) {
+                [self.requestHandlers removeObjectForKey:request];
+            }
+        }
+    }];
     [self removeRequest:request];
 }
 
@@ -222,7 +261,7 @@ NS_ASSUME_NONNULL_BEGIN
     
     BOOL didReceiveDataChunkSelectorExists = [self.delegate respondsToSelector:@selector(dataLoader:didReceiveDataChunk:forResponse:)];
     if (didReceiveDataChunkSelectorExists) {
-        [self executeDelegateBlock: ^{
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didReceiveDataChunk:data forResponse:response];
         }];
     }
@@ -240,7 +279,7 @@ NS_ASSUME_NONNULL_BEGIN
     }
     
     if ([self.delegate respondsToSelector:@selector(dataLoader:didReceiveInitialResponse:)]) {
-        [self executeDelegateBlock: ^{
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didReceiveInitialResponse:response];
         }];
     }
--- a/include/SPTDataLoader/SPTDataLoaderImplementation.h
+++ b/include/SPTDataLoader/SPTDataLoaderImplementation.h
@@ -23,6 +23,7 @@
 @protocol SPTDataLoaderCancellationToken;
 @protocol SPTDataLoaderDelegate;
 @class SPTDataLoaderRequest;
+@class SPTDataLoaderResponse;
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -58,6 +59,18 @@ NS_ASSUME_NONNULL_BEGIN
  @return A cancellation token associated with the request, or `nil` if the request coulnd’t be performed.
  */
 - (nullable id<SPTDataLoaderCancellationToken>)performRequest:(SPTDataLoaderRequest *)request;
+
+/**
+ An alternative performRequest method that takes in a response handler and invokes the handler when the request is complete.
+ This enables a single data loader object to be reused by multiple objects that don't need to implement the delegat protocol
+ and can all submit requests.
+ @discussion The completion handler will be invoked only for terminal states, i.e. when a request succeeded or failed,
+ or when it was canceled.  If the request was cancelled, the reponse parameter of the completion handler will be nil.
+ On success or failure the reponse will be not nil and the completion handler must examine the response in order to
+ determine whether the request succeeded or failed.
+ */
+- (nullable id<SPTDataLoaderCancellationToken>)performRequest:(SPTDataLoaderRequest *)request
+                                               withCompletion:(void (^)(SPTDataLoaderResponse *response))completion;
 
 #pragma mark Cancelling Loads
 
--- a/Tests/SPTDataLoaderTest.m
+++ b/Tests/SPTDataLoaderTest.m
@@ -299,4 +299,55 @@
     XCTAssertNil(cancellationToken, @"The data loader did not release the cancellation token");
 }
 
+- (void)testPerformRequestWithCompletion
+{
+    SPTDataLoaderRequest *request = [SPTDataLoaderRequest new];
+    [self.dataLoader performRequest:request withCompletion:^(SPTDataLoaderResponse *response){}];
+    XCTAssertNotNil(self.requestResponseHandlerDelegate.lastRequestPerformed, @"There should be a valid last request performed");
+}
+
+- (void)testSuccessfulResponseCompletion
+{
+    SPTDataLoaderRequest *request = [SPTDataLoaderRequest new];
+    SPTDataLoaderDelegateMock *delegate = self.dataLoader.delegate;
+    self.dataLoader.delegate = nil;
+    __block SPTDataLoaderResponse *received;
+    [self.dataLoader performRequest:request withCompletion:^(SPTDataLoaderResponse *response){
+        received = response;
+    }];
+    SPTDataLoaderResponse *response = [SPTDataLoaderResponse dataLoaderResponseWithRequest:request response:nil];
+    [self.dataLoader successfulResponse:response];
+    self.dataLoader.delegate = delegate;
+    XCTAssertNotNil(received, @"Response should not be nil");
+}
+
+- (void)testFailureResponseCompletion
+{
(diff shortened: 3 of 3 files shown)


Sources/SPTDataLoader.m
```diff
@@ -62,6 +64,7 @@ - (instancetype)initWithRequestResponseHandlerDelegate:(id<SPTDataLoaderRequestR
         _cancellationTokens = [NSMutableArray new];
         _delegateQueue = dispatch_get_main_queue();
         _requests = [NSMutableArray new];
+        _requestHandlers = [NSMapTable new];
```
dflems: If using `NSMapTable`, use one of the recommended constructors: https://developer.apple.com/documentation/foundation/nsmaptable?language=objc#1661241

Sources/SPTDataLoader.m
```diff
@@ -130,6 +133,26 @@ - (void)removeRequest:(SPTDataLoaderRequest *)request
...
+                                                                                                                 cancelObject:copiedRequest];
+    copiedRequest.cancellationToken = cancellationToken;
+    @synchronized(self.cancellationTokens) {
+        [self.cancellationTokens addObject:cancellationToken];
+    }
+    @synchronized(self.requests) {
+        [self.requests addObject:copiedRequest];
+    }
+    @synchronized(self.requestHandlers) {
+        [self.requestHandlers setObject:completion forKey:request];
```
dflems: Shouldn't this use `copiedRequest`?

Sources/SPTDataLoader.m
```diff
@@ -175,10 +198,19 @@ - (void)successfulResponse:(SPTDataLoaderResponse *)response
...
     }
-    
-    [self executeDelegateBlock: ^{
-        [self.delegate dataLoader:self didReceiveSuccessfulResponse:response];
-    }];
+    if (self.delegate) {
+        [self executeDelegateBlock:^{
+            [self.delegate dataLoader:self didReceiveSuccessfulResponse:response];
+        }];
+    } else {
```
dflems: Handlers should probably be resolved regardless of whether or not a delegate is set

Sources/SPTDataLoader.m
```diff
@@ -175,10 +198,19 @@ - (void)successfulResponse:(SPTDataLoaderResponse *)response
...
+        [self executeDelegateBlock:^{
+            void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:response.request];
+            handler([SPTDataLoaderResult fromResponse:response kind:SPTDataLoaderResultSuccess]);
```
dflems: Always check for `nil` before invoking a block. If `handler == nil`, this will crash.

Sources/SPTDataLoader.m
```diff
@@ -188,11 +220,19 @@ - (void)failedResponse:(SPTDataLoaderResponse *)response
...
-    }];
-
+    if (self.delegate) {
+        [self executeDelegateBlock:^{
+            [self.delegate dataLoader:self didReceiveErrorResponse:response];
+        }];
+    } else {
+        [self executeDelegateBlock:^{
+            void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:response.request];
+            handler([SPTDataLoaderResult fromResponse:response kind:SPTDataLoaderResultFailure]);
```
dflems: Always check for `nil` before invoking a block. If `handler == nil`, this will crash.

Sources/SPTDataLoader.m
```diff
@@ -201,13 +241,19 @@ - (void)cancelledRequest:(SPTDataLoaderRequest *)request
...
-
     if ([self.delegate respondsToSelector:@selector(dataLoader:didCancelRequest:)]) {
-        [self executeDelegateBlock: ^{
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didCancelRequest:request];
         }];
+    } else {
+        [self executeDelegateBlock:^{
+            void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:request];
+            handler([SPTDataLoaderResult fromRequest:request kind:SPTDataLoaderResultCancelled]);
```
dflems: Always check for `nil` before invoking a block. If `handler == nil`, this will crash.

Sources/SPTDataLoader.m
```diff
@@ -240,9 +294,17 @@ - (void)receivedInitialResponse:(SPTDataLoaderResponse *)response
...
     
     if ([self.delegate respondsToSelector:@selector(dataLoader:didReceiveInitialResponse:)]) {
-        [self executeDelegateBlock: ^{
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didReceiveInitialResponse:response];
         }];
+    } else {
+        [self executeDelegateBlock:^{
+            void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:response.request];
+            handler([SPTDataLoaderResult fromResponse:response kind:SPTDataLoaderResultReceivedInitialResponse]);
```
dflems: Always check for `nil` before invoking a block. If `handler == nil`, this will crash.

Sources/SPTDataLoader.m
```diff
@@ -222,9 +268,17 @@ - (void)receivedDataChunk:(NSData *)data forResponse:(SPTDataLoaderResponse *)re
...
     BOOL didReceiveDataChunkSelectorExists = [self.delegate respondsToSelector:@selector(dataLoader:didReceiveDataChunk:forResponse:)];
     if (didReceiveDataChunkSelectorExists) {
-        [self executeDelegateBlock: ^{
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didReceiveDataChunk:data forResponse:response];
         }];
+    } else {
+        [self executeDelegateBlock:^{
+            void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:response.request];
+            handler([SPTDataLoaderResult fromResponse:response withChunk:data]);
```
dflems: Always check for `nil` before invoking a block. If `handler == nil`, this will crash.

Sources/SPTDataLoader.m
```diff
@@ -252,6 +314,14 @@ - (void)requestIsWaitingForConnectivity:(SPTDataLoaderRequest *)request
         [self executeDelegateBlock:^{
             [self.delegate dataLoader:self requestIsWaitingForConnectivity:request];
         }];
+    } else {
+        [self executeDelegateBlock:^{
+            void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:request];
+            handler([SPTDataLoaderResult fromRequest:request kind:SPTDataLoaderResultIsWaitingForConnectivity]);
```
dflems: Always check for `nil` before invoking a block. If `handler == nil`, this will crash.

Sources/SPTDataLoader.m
```diff
@@ -252,6 +314,14 @@ - (void)requestIsWaitingForConnectivity:(SPTDataLoaderRequest *)request
...
+            @synchronized(self.requestHandlers) {
+                [self.requestHandlers removeObjectForKey:request];
+            }
```
dflems: Do you really want to remove the handler here when waiting for connectivity? Then nothing will fire when the requests succeeds or fails

Sources/SPTDataLoaderResult.m
```diff
@@ -0,0 +1,91 @@
...
+#import <SPTDataLoader/SPTDataLoaderResponse.h>
+#import <SPTDataLoader/SPTDataLoaderResult.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation SPTDataLoaderResult
+
++ (instancetype)fromResponse:(SPTDataLoaderResponse *)response kind:(SPTDataLoaderResultKind)kind
+{
+    @synchronized(self.class) {
```
dflems: there is no reason this needs to be synchronized. what shared resource is being accessed?

dflems:
To me it really seems like `performRequest:withCompletion:` is a convenience API for the general case and you don't _NEED_ to support every possible delegate callback with it (chunk received, waiting for connectivity, etc). Just have the block yield a `SPTDataLoaderResponse *` and invoke it on success/failure only.

Sources/SPTDataLoader.m
dmakarov: apparently not.

gitrema: @dmakarov can you elaborate why `copiedRequest` is not needed here?

Sources/SPTDataLoader.m
gitrema: That sounds reasonable.
A check before calling `executeDelegateBlock` can made to see if we have any handlers to execute

Sources/SPTDataLoader.m
gitrema: @dmakarov is this comment still valid? I cannot see the handler being removed

Sources/SPTDataLoader.m
dmakarov: Initially I used the copiedRequest as a key and it didn't work.  Using the actual request is correct.

Sources/SPTDataLoader.m
dmakarov: The handler is no longer removed.  I removed the removing of the handler in this case.

Sources/SPTDataLoader.m
dmakarov: Does this mean both delegate and handler must be invoked, when both exist?

Sources/SPTDataLoader.m
gitrema: I guess because as key it uses the address of the object so you need to original one to retrieve it later.

Sources/SPTDataLoader.m
gitrema: Usually my rule of thumb is "if you ask for weird things you get weird things" I would not expect a client to set both a delegate and block but I'd rather enforce them at API level to not be possible. If we allow them to be set both so the expectation is that both are being called.
The answer to your question, yes, if both exist then both has to be called.

Sources/SPTDataLoader.m
```diff
@@ -240,10 +289,19 @@ - (void)receivedInitialResponse:(SPTDataLoaderResponse *)response
...
+        [self executeDelegateBlock:^{
             [self.delegate dataLoader:self didReceiveInitialResponse:response];
         }];
     }
+    [self executeDelegateBlock:^{
+        void (^handler)(SPTDataLoaderResult *) = [self.requestHandlers objectForKey:response.request];
+        if (handler) {
+            handler([SPTDataLoaderResult fromResponse:response kind:SPTDataLoaderResultReceivedInitialResponse]);
+            @synchronized(self.requestHandlers) {
+                [self.requestHandlers removeObjectForKey:response.request];
```
dflems: I still don't get why this is being done. You are attaching a single callback block but are firing it on _any_ delegate method but you are also removing the callback every time after it is called. None of the subsequent events will fire in the block as you have removed the callback. Only the first one received will ever be fired.

This API also feels sort of clunky and weird (one callback block that is fired many times for many different types of events whether you want them or not). What is the purpose of this?

dflems: If you're going to use a callback for this, it should only be removed for terminal events (request is completely finished).

The naming convention you've gone for is also nonintuitive. The block is called `completion` but it's not really a "completion" block if it's called many times for every type of delegated event. It's almost more of a subscription with how it seems you've defined this API.

gitrema:
LGTM now,
@dflems do you have any further concerns on this?

dflems:
@gitrema I still think this API is clunky and I think the intermediate "result" object is unnecessary. That comment was not addressed.

Sources/SPTDataLoader.m
```diff
@@ -37,6 +37,7 @@ @interface SPTDataLoader () <SPTDataLoaderCancellationTokenDelegate>
 @property (nonatomic, strong, readonly) NSMutableArray<id<SPTDataLoaderCancellationToken>> *cancellationTokens;
 @property (nonatomic, strong, readonly) NSMutableArray<SPTDataLoaderRequest *> *requests;
 @property (nonatomic, strong, readonly) id<SPTDataLoaderCancellationTokenFactory> cancellationTokenFactory;
+@property (nonatomic, strong) NSMapTable *requestHandlers;
```
themackworth: should be readonly?

Sources/SPTDataLoader.m
```diff
@@ -130,6 +132,26 @@ - (void)removeRequest:(SPTDataLoaderRequest *)request
...
+    id<SPTDataLoaderCancellationToken> cancellationToken = [self.cancellationTokenFactory createCancellationTokenWithDelegate:self
+                                                                                                                 cancelObject:copiedRequest];
+    copiedRequest.cancellationToken = cancellationToken;
+    @synchronized(self.cancellationTokens) {
+        [self.cancellationTokens addObject:cancellationToken];
+    }
+    @synchronized(self.requests) {
+        [self.requests addObject:copiedRequest];
+    }
+    @synchronized(self.requestHandlers) {
```
themackworth: Did you consider trying to put the handler block on the request itself in some way (like userInfo).  There would be pros and cons to that, for example, it might be a little weird to modify the request in that way and have the blocks existence leak out of the response object, but I am curious if that can would be outweighed by not having to use synchronized.

include/SPTDataLoader/SPTDataLoaderImplementation.h
```diff
@@ -59,6 +60,18 @@ NS_ASSUME_NONNULL_BEGIN
...
+ An alternative performRequest method that takes in a response handler and invokes the handler when the request is complete.
+ This enables a single data loader object to be reused by multiple objects that don't need to implement the delegat protocol
+ and can all submit requests.
+ @discussion The completion handler will be invoked only for terminal states, i.e. when a request succeeded or failed,
+ or when it was canceled.  If the request was cancelled, the reponse parameter of the completion handler will be nil.
+ On success or failure the reponse will be not nil and the completion handler must examine the response in order to
+ determine whether the request succeeded or failed.
+ */
+- (nullable id<SPTDataLoaderCancellationToken>)performRequest:(SPTDataLoaderRequest *)request
+                                               withCompletion:(void (^)(SPTDataLoaderResponse *response))completion;
```
themackworth: Only including the response and not the potential error in the block params seems a little unorthodox.

themackworth:
I always use blocks in userInfo when I use SPTDataLoader, so I am pro block, but the docs do have a section titled "Why no block interface", and I am not sure this API addresses those concerns (making error handling obvious and straightforward).  I am not saying we can't move forward with blocks or this PR, but I just wanted to throw that out there.  At the very least, the docs won't be accurate after this change.

Sources/SPTDataLoader.m
dmakarov: I believe we discussed this internally and decided not to attach completion blocks to requests in the userInfo field.

erikprice:
I know this is a relatively small change, and overall I think a good one, but a couple of [code smells](https://en.wikipedia.org/wiki/Code_smell) stand out in the design:

* It's weird to have an API that works with both blocks _and_ delegates. I'm sure there's prior art on that, but it would be simpler for the API to allow clients to use one or the other, but not (either accidentally or intentionally) both.
* It feels like an antipattern to have "nearly-parallel" data structures holding requests, request handlers, and cancellation tokens – seems like there ought to be some way to unify them into a single data structure, such as by just storing a single list of "request contexts" that each hold references to all of the disposable objects associated with a call to `-performRequest:`. The proliferation of these data structure instance variables introduces accidental complexity in the form of additional state management and synchronization. More things can go wrong. It also just makes it harder to reason about the code.

I'd like to see us solve problems with composition where possible. One way you could avoid the above issues and also implement your enhancement without intruding on the existing SPTDataLoader API and implementation would be to compose SPTDataLoader behind a simple facade that exposes the functionality for performing requests with success and failure handlers, and which manages the lifetimes of the objects. The facade could be initialized with a SPTDataLoaderFactory and offer only the `-performRequest:completion:` method proposed in this PR, and create a new SPTDataLoader instance every time this method is called. Whenever a SPTDataLoader's delegate is notified of a successful or failed request, the facade would call the appropriate request handler and then remove all of the state associated with that request (the request, request handlers, and the SPTDataLoader doing the actual work).

Something like

```objc
@implementation SPTClosureBasedDataLoader  // TODO: needs a better name ;-)

- (instancetype)initWithDataLoaderFactory:(SPTDataLoaderFactory *)factory
{
    if (self) {
        _factory = factory;
    }

    return self;
}

- (id<SPTDataLoaderCancellationToken>)performRequest:(SPTDataLoaderRequest *)request
                                      successHandler:(void (^)(SPTDataLoaderResponse *response))successHandler
                                      failureHandler:(void (^)(SPTDataLoaderResponse *response))failureHandler
{
    SPTDataLoader *loader = [self.factory createDataLoader];
    loader.delegate = self;
    SPTDataLoaderRequestCall *call =
        [SPTDataLoaderRequestCall callWithRequest:request
                                           loader:loader
                                   successHandler:successHandler
                                   failureHandler:failureHandler];
    [self.calls addObject:call];

    return [loader performRequest:request];
}

- (void)dataLoader:(SPTDataLoader *)dataLoader didReceiveSuccessfulResponse:(SPTDataLoaderResponse *)response
{
    // Look up the SPTDataLoaderRequestCall corresponding to response.request,
    // then invoke its successHandler.
    // Finally, clean up by removing the entire call from `self.calls`.
}

// (etc for didReceiveErrorResponse)

@end
```

I'm glossing over the details, and of course there are other ways to skin this cat, but a composition-based approach like this reduces the impact on SPTDataLoader's existing code and preserves the separation of concerns with respect to a delegate-based design vs a closure-based design, providing a dedicated path for clients that don't want to use a delegate-based API.

rastersize:
I’m also very late to this PR, sorry about that. If we want to introduce this change we should also revisit the readme’s section that explains why SPTDataLoader doesn’t have a block based API. See https://github.com/spotify/SPTDataLoader#why-no-block-interface

include/SPTDataLoader/SPTDataLoaderImplementation.h
dflems: The error is a property on the response (when you get the error delegate callback you also don't explicitly get an error object IIRC)

themackworth: yeah, that's a good point.

PR #188, opened 2020-06-08
Add resolvedURL property to response object

dflems:
After following redirects, the value in `-[NSURLResponse URL]` may be different from that of the request. There are instances where we want to capture this.


Sources/SPTDataLoaderResponse.m
```diff
@@ -52,7 +51,6 @@ - (instancetype)initWithRequest:(SPTDataLoaderRequest *)request response:(nullab
     self = [super init];
     if (self) {
         _request = request;
-        _response = response;
```
dflems, line 55: This was only being retained for the `description` method. Everything else was copied out of it.

PR #189, opened 2020-06-19
Fix issue that was making the DataLoaderRateLimiter delaying requests unnecessarily if the system clock was moved backwards

tanzolone:
**Context**
The `SPTDataLoaderRateLimiter` has a method to compute when is earliest time a new request can be executed: `earliestTimeUntilRequestCanBeExecuted`. This includes a throttle to avoid that too many requests are sent in a very short period of time. The object is initialised with a `requestsPerSecond` parameter which then determines the minimum waiting time between two requests: `1 / requestsPerSecond`.

The lines that apply this logic as follows:
```
CFAbsoluteTime deltaTime = currentTime - lastExecution;
CFAbsoluteTime cutoffTime = 1.0 / requestsPerSecond;
CFAbsoluteTime timeInterval = cutoffTime - deltaTime;
if (timeInterval < 0.0) {
  timeInterval = 0.0;
}
return timeInterval;
```

**Problem**
In normal conditions `currentTime` is always going to be larger than `lastExecution`. However there is an edge-case: if the user changes the system time (e.g. move clock backwards in iOS settings) then `currentTime` would become less than `lastExecution` and the above `deltaTime` would be unexpectedly negative. With the current logic this would affect the computation of `timeInterval` which would become ~ as large as - `deltaTime`. In other words if the user moves the clock 24h backwards we would not allow any requests for ~24h.

**Suggested Solution**
Ideally we should use a Monotonic clock (e.g. `mach_absolute_time()`) but this would require some careful consideration and major refactoring as `CFAbsoluteTimeGetCurrent()` is the reference time almost everywhere in the framework.

As I don't see any other reasons for `currentTime` being less than `lastExecution` other than the scenario described above, I think we should simply add the following check and allow requests immediately if that happens:
```
if (deltaTime < 0) {
  // If currentTime < lastExecution the system clock must have been moved backwards
  // We should execute the request immediately to allow the RateLimiter to resume working as expected
  return 0;
}
```


tanzolone:
Looking into a different option using system uptime instead

PR #191, opened 2020-07-24
Data loader: add a thin wrapper with a simple block API

marmelroy:
Looking at usages of Data Loader within Spotify and specifically in Swift features, many engineers are looking for a block based API to handle the common cases of data loading. 

This idea was already explored in https://github.com/spotify/SPTDataLoader/pull/186 but seemed to have been abandoned. 

This PR builds on top of the previous discussion and creates a thin wrapper that provides a block based API. The API design should also nudge users to handle the error case.

@dflems @erikprice @themackworth @JensAyton @rastersize

### DIFF
--- a/Sources/SPTDataLoaderBlockWrapper.m
+++ b/Sources/SPTDataLoaderBlockWrapper.m
@@ -18,7 +18,7 @@
  specific language governing permissions and limitations
  under the License.
  */
-#import "SPTDataLoaderBlockWrapper.h"
+#import <SPTDataLoader/SPTDataLoaderBlockWrapper.h>
 #import "SPTDataLoader.h"
 
 NS_ASSUME_NONNULL_BEGIN
--- a/Sources/SPTDataLoaderBlockWrapper.m
+++ b/Sources/SPTDataLoaderBlockWrapper.m
@@ -1,0 +1,87 @@
+/*
+ Copyright (c) 2015-2020 Spotify AB.
+
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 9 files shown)


include/SPTDataLoader/SPTDataLoaderBlockWrapper.h
```diff
@@ -0,0 +1,51 @@
...
+ under the License.
+ */
+#import <Foundation/Foundation.h>
+
+@class SPTDataLoader;
+@class SPTDataLoaderRequest;
+@class SPTDataLoaderResponse;
+@protocol SPTDataLoaderCancellationToken;
+
+typedef void (^DataLoaderBlockCompletion)(SPTDataLoaderResponse * _Nonnull response, NSError *_Nullable error);
```
dflems, line 28: `SPTDataLoaderBlockCompletion`, please prefix.

Sources/SPTDataLoaderBlockWrapper.m
```diff
@@ -0,0 +1,87 @@
...
+        _dataLoader = dataLoader;
+        dataLoader.delegate = self;
+        _completionHandlers = [NSMapTable strongToWeakObjectsMapTable];
+    }
+    return self;
+}
+
+- (nullable id<SPTDataLoaderCancellationToken>)performRequest:(SPTDataLoaderRequest *)request completion:(DataLoaderBlockCompletion)completion
+{
+    NSString *sourceIdentifier = request.sourceIdentifier;
```
dflems, line 48: No guarantee that `sourceIdentifier` is sufficiently unique per-request. Drop a UUID in the `userInfo` dictionary of the request object with a specific key name.

PR #194, opened 2020-09-01
Introduce Swift-only request APIs

kmcbride:
These changes take the idea behind [SPTDataLoaderBlockWrapper.m](https://github.com/spotify/SPTDataLoader/blob/master/Sources/SPTDataLoaderBlockWrapper.m) and expand upon it to provide protocol-oriented APIs that better utilize Swift language features.

#### Example Usage
```swift
struct Album: Decodable {
    let title: String
    let artistName: String
}

let dataLoader = dataLoaderFactory.makeDataLoader()
let request = SPTDataLoaderRequest(url: albumURL, sourceIdentifier: "album-feature")

dataLoader.request(request) { (response: Response<Album, Error>) in
    switch response.result {
    case .success(let album):
        print("Loaded album: \(album.title) by \(album.artistName)"
    case .failure(let error):
        print("Failed to load album: \(error)")
    }
}
```

#### Adding Custom Response Serializer
```swift
import SwiftProtobuf

struct ProtobufResponseSerializer<Message: SwiftProtobuf.Message>: ResponseSerializer {
    func serialize(response: SPTDataLoaderResponse) throws -> Message {
        if let error = response.error {
            throw error
        }

        return try Message(serializedData: response.body ?? Data())
    }
}

extension DataLoader {
    @discardableResult
    func request<Message: SwiftProtobuf.Message>(
        _ request: SPTDataLoaderRequest,
        completionHandler: @escaping (Response<Message, Error>) -> Void
    ) -> SPTDataLoaderCancellationToken? {
        let serializer = ProtobufResponseSerializer<Message>()

        return self.request(request, serializer: serializer, completionHandler: completionHandler)
    }
}
```


rastersize:
@kmcbride One opportunity of improvement I see with the request-handler closure is to expose the HTTP response. So that the closure can for example look at the HTTP status code. I find that quite useful form time to time. For example when distinguishing between a network error and a server error. Or when deciding how to handle an error, like differentiating between a HTTP 403 and 500 status code.

Also, have you considered adding a method for parsing server errors? E.g. something like:

```swift
dataLoader
    .validate(ServerError.self)
    .request(request) { (response: Response<Album, ServerError>) in ... }

// Or
dataLoader.request(validateUsing: Validator.self, request) { (response: Response<Album, ServerError>) in ... }

// Or
dataLoader.request(validateUsing: validator, request) { (response: Response<Album, ServerError>) in ... }
```

kmcbride:
@rastersize I like the idea of being able to chain a validator and I did explore it at one point, but it seemed like it would require refactoring internals beyond what could be provided by a simple wrapper. Including it as a parameter to the request method is a possibility, but would cause a bit of API bloat due to the inability to use default values in the protocol. In the interim, I've added an additional `request(…)` function that provides the raw `SPTDataLoaderResponse` to the completion handler. It's also possible to provide a custom serializer that makes use of the response object to determine what result to provide.

rastersize:
@kmcbride @dflems Btw. since we’re adding Swift overlays to SPTDataLoader, should we also add our `.swiftlint.yml` config and run it in CI?

Sources/SPTDataLoaderSwift/DataLoaderWrapper.swift
```diff
@@ -0,0 +1,134 @@
...
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ */
+
+import Foundation
+
+/// A wrapper around `SPTDataLoader` that allows it to be used with a
+/// block-based API without having to pass closures through userInfo.
```
rastersize, line 25: nit:
```suggestion
/// closure-based API without having to pass closures through `userInfo`.
```

Sources/SPTDataLoaderSwift/DataLoaderWrapper.swift
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public func dataLoader(_ dataLoader: SPTDataLoader, didReceiveErrorResponse response: SPTDataLoaderResponse) {
+        if let handler = responseHandlers.removeValue(forKey: response.request.uniqueIdentifier) {
+            handler(response)
+        }
+    }
+
+    public func dataLoader(_ dataLoader: SPTDataLoader, didCancel request: SPTDataLoaderRequest) {
+        responseHandlers.removeValue(forKey: request.uniqueIdentifier)
```
rastersize, line 132: Is there no need for the handling closure to be notified of when a request is cancelled?

Sources/SPTDataLoaderSwift/SPTDataLoaderFactory+Convenience.swift
```diff
@@ -0,0 +1,35 @@
...
+import Foundation
+
+public extension SPTDataLoaderFactory {
+    /// Convenience method for creating a Swift `DataLoader`.
+    func makeDataLoader() -> DataLoader {
+        let sptDataLoader = createDataLoader()
+        let dataLoaderWrapper = DataLoaderWrapper(sptDataLoader: sptDataLoader)
+
+        sptDataLoader.delegate = dataLoaderWrapper
+        sptDataLoader.delegateQueue = .global()
```
rastersize, line 31: Would it make sense to allow this to be customizable? E.g. via a default parameter.

Sources/SPTDataLoaderSwift/DataLoaderWrapper.swift
kmcbride: The need is likely out there, but for the sake of simplicity — while still handling the vast majority of use cases I've seen in our apps — it's been omitted for now.

rastersize: Sounds reasonable!

dflems:
@cedercra The codecov file only has _ignores_ not paths to look for code. Something is broken on their end.

@kmcbride If you give me write access to your fork, I can fiddle with this locally.

rastersize:
My comment was a bit confusing, I meant to add the new tests directory to the exclude list. I was hoping that would get Codecov to pick things up. If nothing else it should get rid of the warnings added to the test files.

kmcbride:
@rastersize @dflems I've somewhat fixed the codecov issues:
1. I corrected the `codecov.yml` because the tests directory stopped being ignored [quite a while ago](https://github.com/spotify/SPTDataLoader/pull/171) 🙂
2. I ran the script locally in order to upload the [correct coverage](https://codecov.io/gh/spotify/SPTDataLoader/pull/194/tree/Sources) — for some reason the line hit counts are reported as 0 from the CI build

rastersize (approved):

PR #195, opened 2020-09-02
Accept-Language: deduplicate preferred localizations

erikprice:
Anecdotally, it seems that NSBundle.preferredLocalizations can return
a list containing duplicate values.


rastersize (approved):
Weird 😶

dflems:
Very weird!

PR #196, opened 2020-09-08
Add an option to use background session.

aniakorolova:
**What has changed**
I added a new initializers to the session selector and the data loader service to enable background networking.

**Why this was changed**
Using background sessions together with download tasks is necessary for networking on watchOS. See details [here](https://developer.apple.com/documentation/watchkit/keeping_your_watchos_content_up_to_date/making_background_requests?language=objc).

### DIFF
--- a/Sources/SPTDataLoaderServiceSessionSelector.m
+++ b/Sources/SPTDataLoaderServiceSessionSelector.m
@@ -27,6 +27,8 @@ NS_ASSUME_NONNULL_BEGIN
 @interface SPTDataLoaderServiceDefaultSessionSelector ()
 
 @property (nonatomic, strong, readonly) NSURLSessionConfiguration *configuration;
+@property (nonatomic, strong, readonly, nullable) NSURLSessionConfiguration *backgroundConfiguration;
+@property (nonatomic, strong, readwrite, nullable) NSURLSession *backgroundSession;
 @property (nonatomic, weak, readonly) id<NSURLSessionDelegate> delegate;
 @property (nonatomic, strong, readonly) NSOperationQueue *delegateQueue;
 
@@ -37,6 +39,19 @@ NS_ASSUME_NONNULL_BEGIN
 {
     NSURLSession *_nonWaitingSession;
     NSURLSession *_waitingSession;
+}
+
+- (instancetype)initWithConfiguration:(NSURLSessionConfiguration *)configuration
+              backgroundConfiguration:(NSURLSessionConfiguration *)backgroundConfiguration
+                             delegate:(id<NSURLSessionDelegate>)delegate
+                        delegateQueue:(NSOperationQueue *)delegateQueue
+{
+    self = [self initWithConfiguration:configuration delegate:delegate delegateQueue:delegateQueue];
+    if (self) {
+        _backgroundConfiguration = [backgroundConfiguration copy];
+    }
+
+    return self;
 }
 
 - (instancetype)initWithConfiguration:(NSURLSessionConfiguration *)configuration
@@ -56,6 +71,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSURLSession *)URLSessionForRequest:(SPTDataLoaderRequest *)request
 {
+    if (request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
+        NSURLSession *backgroundSession = self.backgroundSession;
+        if (backgroundSession) {
+            return backgroundSession;
+        }
+    }
+
     if (request.waitsForConnectivity) {
         return self.waitingSession;
     } else {
@@ -69,6 +91,17 @@ NS_ASSUME_NONNULL_BEGIN
         _waitingSession = [self createWaitingSession];
     }
     return _waitingSession;
+}
+
+- (nullable NSURLSession *)backgroundSession
+{
+    NSURLSessionConfiguration *backgroundSessionConfiguration = self.backgroundConfiguration;
+    if (!_backgroundSession && backgroundSessionConfiguration) {
+        _backgroundSession = [NSURLSession sessionWithConfiguration:backgroundSessionConfiguration
+                                                           delegate:self.delegate
+                                                      delegateQueue:self.delegateQueue];
+    }
+    return _backgroundSession;
 }
 
 - (NSURLSession *)nonWaitingSession
@@ -98,6 +131,7 @@ NS_ASSUME_NONNULL_BEGIN
 {
     [self.waitingSession invalidateAndCancel];
     [self.nonWaitingSession invalidateAndCancel];
+    [self.backgroundSession invalidateAndCancel];
 }
 
 @end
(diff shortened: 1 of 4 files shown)


Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -27,6 +27,8 @@ Licensed to the Apache Software Foundation (ASF) under one
 @interface SPTDataLoaderServiceDefaultSessionSelector ()
 
 @property (nonatomic, strong, readonly) NSURLSessionConfiguration *configuration;
+@property (nonatomic, strong, readonly, nullable) NSURLSessionConfiguration *backgroundConfiguration;
```
gitrema, line 30: what makes the `backgroundConfiguration` special?
Cannot the `configuration` be used by just changing the needed fields?

Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -27,6 +27,8 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+@property (nonatomic, strong, readwrite, nullable) NSURLSession *backgroundSession;
```
gitrema, line 31: I guess this can be move where the other Session are. Line `40` and `41`

Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -56,6 +71,13 @@ - (instancetype)initWithConfiguration:(NSURLSessionConfiguration *)configuration
 
 - (NSURLSession *)URLSessionForRequest:(SPTDataLoaderRequest *)request
 {
+    if (request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
+        NSURLSession *backgroundSession = self.backgroundSession;
```
gitrema, line 75: it can be assumed it background session is always there if it is created internally instead of being injected at `init`

Sources/SPTDataLoaderServiceSessionSelector.m
aniakorolova: A background configuration object should be created using this method
https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1407496-backgroundsessionconfigurationwi

Sources/SPTDataLoaderServiceSessionSelector.m
aniakorolova: I could move it there, but what would be the benefit of that?

Sources/SPTDataLoaderServiceSessionSelector.m
gitrema: ok so can it be created internally, like it is done in `createWaitingSession`, instead of having it injected?

Sources/SPTDataLoaderServiceSessionSelector.m
aniakorolova: In this setup, the background session will only be created if a configuration object for it is provided at init. It looks safer to me to provide a new initializer instead of changing this logic for everyone who's already using data loader.

Sources/SPTDataLoaderServiceSessionSelector.m
aniakorolova: Yeah, it can be created internally

include/SPTDataLoader/SPTDataLoaderService.h
```diff
@@ -66,6 +66,18 @@ NS_ASSUME_NONNULL_BEGIN
                                        rateLimiter:(nullable SPTDataLoaderRateLimiter *)rateLimiter
                                           resolver:(nullable SPTDataLoaderResolver *)resolver;
 
+/**
+ Convenience constructro with background networking settings.
```
dflems, line 70: "constructor"

constructro is a cool name for a robot though 🤖

Sources/SPTDataLoaderServiceSessionSelector.m
aniakorolova: I should make it readonly, actually, so I can add an instance var there instead of synthesizing.

aniakorolova:
I removed the background session configuration from the parameters in `SPTDataLoader.dataLoaderWith...`. This seems to make the interface confusing though.

Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -98,6 +132,7 @@ - (void)invalidateAndCancel
 {
     [self.waitingSession invalidateAndCancel];
     [self.nonWaitingSession invalidateAndCancel];
+    [self.backgroundSession invalidateAndCancel];
```
dflems, line 135: This should use the ivar, not the property. If up until this point, the background session had not been created, it will be created just to invalidate it.

kmcbride (commented):
Edit: On second thought, wouldn't it be better for the API user to pass their own background configuration object to the existing initializer? No changes are necessary and it allows for adjusting the properties on the configuration if needed.

---

The additional initializer seems unnecessary and somewhat confusing because it doesn't actually expect a background configuration, it simply calls through to the existing initializer and creates a background configuration that doesn't rely on any of the method parameters.

Couldn't the same result be achieved transparently to the consumer by lazily initializing the `backgroundSession` property in the same way `nonWaitingSession` is?

```objective-c
- (NSURLSession *)URLSessionForRequest:(SPTDataLoaderRequest *)request
{
    if (request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
        return self.backgroundSession;
    }

    if (request.waitsForConnectivity) {
        return self.waitingSession;
    }

    return self.nonWaitingSession;
}

- (NSURLSession *)backgroundSession
{
    if (_backgroundSession == nil) {
        NSURLSessionConfiguration *configuration =
            [NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:SPTDataLoaderBackgroundSessionConfigurationIdentifier];
        _backgroundSession = [NSURLSession sessionWithConfiguration:configuration
                                                           delegate:self.delegate
                                                      delegateQueue:self.delegateQueue];
    }

    return _backgroundSession;
}
```

Sources/SPTDataLoaderServiceSessionSelector.m
gitrema: Not really. It opens up to bugs and clutter the API.
I actually I think the missing of an internal background session is kind of a bug if background tasks are allowed

gitrema:
> Edit: On second thought, wouldn't it be better for the API user to pass their own background configuration object to the existing initializer? No changes are necessary and it allows for adjusting the properties on the configuration if needed.
> 

I'd avoid to pass any new configuration also. The SPTDataLoader already allows background tasks but there is no internal backgroundSession and it can easily be created internally.

Sources/SPTDataLoaderServiceSessionSelector.m
aniakorolova: Oh that's true, thanks

aniakorolova:
I'll close this for now.

aniakorolova:
Thank you for the comments @gitrema @kmcbride @dflems, I opened another PR here https://github.com/spotify/SPTDataLoader/pull/197
which creates a background session internally in the session selector.

kmcbride:
> I'd avoid to pass any new configuration also. The SPTDataLoader already allows background tasks but there is no internal backgroundSession and it can easily be created internally.

@gitrema fair enough. My main concern would be that creating the configuration internally makes it impossible for the framework user to set properties that are relevant to [background transfers](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration#topics) ("Supporting Background Transfers" section) and also prevents their app from [conditionally handling](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622941-application?language=objc) background sessions. So, if anything, a change like this would break functionality that is currently possible.

PR #197, opened 2020-09-09
Session selector: Use a background session for background requests

aniakorolova:
Another take on background downloads.

This will create a background session in session selector to use for all requests with background policy set to `SPTDataLoaderRequestBackgroundPolicyAlways`.

### DIFF
--- a/Sources/SPTDataLoaderServiceSessionSelector.m
+++ b/Sources/SPTDataLoaderServiceSessionSelector.m
@@ -24,9 +24,13 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+static NSString *const SPTDataLoaderBackgroundSessionConfigurationIdentifier =
+    @"SPTDataLoaderBackgroundSessionConfigurationIdentifier";
+
 @interface SPTDataLoaderServiceDefaultSessionSelector ()
 
 @property (nonatomic, strong, readonly) NSURLSessionConfiguration *configuration;
+@property (nonatomic, strong, readonly) NSURLSession *backgroundSession;
 @property (nonatomic, weak, readonly) id<NSURLSessionDelegate> delegate;
 @property (nonatomic, strong, readonly) NSOperationQueue *delegateQueue;
 
@@ -37,6 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 {
     NSURLSession *_nonWaitingSession;
     NSURLSession *_waitingSession;
+    NSURLSession *_backgroundSession;
 }
 
 - (instancetype)initWithConfiguration:(NSURLSessionConfiguration *)configuration
@@ -56,6 +61,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSURLSession *)URLSessionForRequest:(SPTDataLoaderRequest *)request
 {
+    if (request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
+        return self.backgroundSession;
+    }
+
     if (request.waitsForConnectivity) {
         return self.waitingSession;
     } else {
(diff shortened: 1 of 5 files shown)


Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -71,6 +80,18 @@ - (NSURLSession *)waitingSession
     return _waitingSession;
 }
 
+- (NSURLSession *)backgroundSession
+{
+    if (!_backgroundSession) {
```
gitrema, line 85: to be consistent with the style in this file, would it be possible to be explicit using `== nil` ?

Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -71,6 +80,18 @@ - (NSURLSession *)waitingSession
...
+        NSURLSessionConfiguration *backgroundSessionConfiguration = [NSURLSessionConfiguration
```
gitrema, line 86: again, just for the sake of being consistent in style, would it be possible to add a `createBackgroundSession` method like the `createWaitingSession` one?

Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -96,8 +117,9 @@ - (NSURLSession *)createWaitingSession
 
 - (void)invalidateAndCancel
 {
-    [self.waitingSession invalidateAndCancel];
-    [self.nonWaitingSession invalidateAndCancel];
```
gitrema, line 100: if possible it would be better to have those changes in a different commit, it is ok if they are part of this PR

Sources/SPTDataLoaderServiceSessionSelector.m
```diff
@@ -24,9 +24,13 @@ Licensed to the Apache Software Foundation (ASF) under one
 
 NS_ASSUME_NONNULL_BEGIN
 
+static NSString *const SPTDataLoaderBackgroundSessionConfigurationIdentifier =
+    @"SPTDataLoaderBackgroundSessionConfigurationIdentifier";
+
 @interface SPTDataLoaderServiceDefaultSessionSelector ()
 
 @property (nonatomic, strong, readonly) NSURLSessionConfiguration *configuration;
+@property (nonatomic, strong, readonly) NSURLSession *backgroundSession;
```
gitrema, line 33: is that needed?

aniakorolova: I think it makes sense to declare a property if we're going to use dot notation for accessing it. But up to you, I can remove it as well.

Sources/SPTDataLoaderServiceSessionSelector.m
dflems: I think this is a good addition. If the sessions haven't been created yet, there's no reason to create them just to invalidate them.

Sources/SPTDataLoaderServiceSessionSelector.m
gitrema: Not a super expert for the Obj-C language so I might be wrong.
wouldn't `self.backgroundSession` call the `backgroundSession` method that was added?

That is what `self.waitingSession` does and there is no property for that. That is why I am asking

aniakorolova: Yes, it will work the same way. Just a code clarity thing.

gitrema: It is not super important but in my opinion to conform to the same style in the file, either this one is not introduced or the other ones are added also.

aniakorolova:
@dflems how do I hint the codecov bot to ignore the coverage for the test file?

PR #200, opened 2020-09-19
Swift: Refine request API

kmcbride:
This follow-up to https://github.com/spotify/SPTDataLoader/pull/194 incorporates some of the suggestions to support requests with chainable validators and handlers. Redundant APIs have been removed in favor of their more capable variations.

#### Example usage
```swift
dataLoader
    // Creating a request
    .request(albumURL, sourceIdentifier: "album-page")
    // Modifying the request properties
    .modify { request in
        request.body = albumData
        request.method = .patch
        request.addValue("application/json", forHeader: "Accept")
    }
    // Validating the response
    .validate { response in
        guard response.statusCode.rawValue == 200 else {
            throw ValidationError.badStatus(code: response.statusCode.rawValue)
        }
    }
    // Executing the request and serializing the response
    .responseDecodable { response in
        eventConsumer(.albumLoaded(result: response.result))
    }
```


rastersize (commented):
A few notes, unfortunately I didn’t have time to finish my review the other day and I won’t have much additional time until next week.

Sources/SPTDataLoaderSwift/DataLoader.swift
```diff
@@ -23,6 +23,11 @@ import Foundation
 
 /// A protocol that provides data request handling.
 public protocol DataLoader {
+    /// Creates a request that can be used to retrieve the contents of a URL.
+    /// - Parameter url: The `URL` for the request.
+    /// - Parameter sourceIdentifier: The identifier for the request source. May be `nil`.
+    func request(_ url: URL, sourceIdentifier: String?) -> Request
```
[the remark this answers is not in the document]
kmcbride: Yeah, this was something I had originally overlooked but later pushed a fix for (possibly after you started your review). There is now a [modify](https://github.com/kmcbride/SPTDataLoader/commit/2511d700a0069138ba06f7af205f9581fb8b0e36#diff-15b566be2565980a08ab01b729f0c3bcR134-R147) method that can be used to mutate the underlying request instance.

eg:
```swift
request.modify { request in
    request.body = albumData
    request.method = .patch
    request.addValue("application/json", forHeader: "Accept")
}
```

Sources/SPTDataLoaderSwift/DataLoaderWrapper.swift
```diff
@@ -117,18 +130,27 @@ extension DataLoaderWrapper: DataLoader {
 
 extension DataLoaderWrapper: SPTDataLoaderDelegate {
     func dataLoader(_ dataLoader: SPTDataLoader, didReceiveSuccessfulResponse response: SPTDataLoaderResponse) {
+        if let request = requests.removeValue(forKey: response.request.uniqueIdentifier) {
```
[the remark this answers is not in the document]
kmcbride: Good point. I've added a commit to synchronize access.

Sources/SPTDataLoaderSwift/DataLoaderError.swift
```diff
@@ -0,0 +1,32 @@
...
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ */
+
+/// An error that occurs during request execution.
+public enum RequestError: Error {
```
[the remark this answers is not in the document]
kmcbride: I was thinking the number of cases might increase over time, so it's partially for future-proofing. Does the struct offer benefits I might be overlooking?

quentar:
Thanks for locking for cancellation and other states, should behaviour with rapid fire requests :-)

Sources/SPTDataLoaderSwift/DataLoader.swift
```diff
@@ -23,59 +23,23 @@ import Foundation
...
-    /// - Returns: A token capable of cancelling the request, or `nil` if the request is invalid.
-    @discardableResult
-    func request(
-        _ request: SPTDataLoaderRequest,
-        completionHandler: @escaping (Response<Data?, Error>) -> Void
-    ) -> SPTDataLoaderCancellationToken?
+    /// Creates a request that can be used to retrieve the contents of a URL.
+    /// - Parameter url: The `URL` for the request.
+    /// - Parameter sourceIdentifier: The identifier for the request source. May be `nil`.
+    func request(_ url: URL, sourceIdentifier: String?) -> Request
```
dflems, line 29: Is this a factory method for a `Request` object or does it perform a request? The comment block doesn't make this clear.

Sources/SPTDataLoaderSwift/Request.swift
```diff
@@ -0,0 +1,287 @@
...
+            default:
+                break
+            }
+        }
+    }
+
+    func addResponseHandler(_ responseHandler: @escaping ResponseHandler) {
+        var result: Result<SPTDataLoaderResponse, Error>?
+
+        accessLock.sync {
```
dflems, line 71: Does this lock need to be on this entire switch statement? Seems like you only need it on the sections where you modify `responseHandlers`. Always best to keep the critical sections as short as possible.

Sources/SPTDataLoaderSwift/Request.swift
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        result.map { result in responseHandler(result) }
+    }
+
+    func processResponse(_ response: SPTDataLoaderResponse) {
+        var error = response.error
+        var handlers: [ResponseHandler] = []
+
+        accessLock.sync {
```
dflems, line 99: same here, you should only lock on things that are mutated that are owned by this object

Sources/SPTDataLoaderSwift/Request.swift
kmcbride: I believe this does need to encompass the entire switch because `state` also needs exclusive access when being read/written.

Sources/SPTDataLoaderSwift/Request.swift
kmcbride: As above, I believe this whole section is critical because of the read + mutate access to `state`, `responseHandlers`, and `responseValidators `.

Sources/SPTDataLoaderSwift/DataLoader.swift
kmcbride: Good point. The comment should specify that it creates a `Request` not a request. Will update.

dflems (approved):

PR #206, opened 2021-05-31
Swift: Add active request convenience methods

kmcbride:
Provides convenience methods for obtaining and cancelling active requests, similar to those of the underlying `SPTDataLoader`.

cc @zvonicek


zvonicek (approved):
Great job, thanks for implementing this.

PR #208, opened 2021-10-31
Create Swift package

kmcbride:
Not sure how much interest there even is for this, but figure it wouldn't hurt to include 🤷


dflems:
Dunno about all the symlinks. If we're gonna do that we might as well fiddle with the repo structure.

kmcbride:
@dflems Fair. Went ahead and reorganized the structure — should be working the same as before 🤞

PR #215, opened 2022-03-01
Swift: Create wrappers for Combine & Concurrency

kmcbride:
Introduces types to support Combine and Swift Concurrency usage.

### Combine
```swift
dataLoader.request(modelURL, sourceIdentifier: "model-page")
    .serializablePublisher(serializer: protobufSerializer)
    .sink { response in completionHandler(response.result) }
    .store(in: &cancellables)
```
```swift
dataLoader.request(imageURL, sourceIdentifier: "album-page")
    .dataPublisher()
    .valuePublisher()
    .sink(
        receiveCompletion: { completion in
            if case .failure(let error) = completion {
                errorHandler(error)
            }
        },
        receiveValue: { imageData in
           completionHandler(imageData)
        }
    )
```

### Swift Concurrency
```swift
let request = dataLoader.request(modelURL, sourceIdentifier: "model-page")
let modelTask = request.serializableTask(serializer: protobufSerializer)
let modelResponse = await modelTask.response
```
```swift
let imageTask = dataLoader.request(imageURL, sourceIdentifier: "album-page").dataTask()
let imageData = try await imageTask.value
```


Sources/SPTDataLoaderSwift/Request+Concurrency.swift
```diff
@@ -0,0 +1,120 @@
...
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+public struct ResponseTask<Value> {
+    private let task: Task<Response<Value, Error>, Never>
+
+    fileprivate init(
+        request: Request,
+        continuation: @escaping (CheckedContinuation<Response<Value, Error>, Never>) -> Void
+    ) {
+        self.task = Task {
+            await withTaskCancellationHandler(
```
gitrema, line 53: can we limit the use of `await/async` to iOS 15 only? Even though there is a backward compatibility library up to iOS 13. It is not clear how well it performs or what bugs might be covering under the surface at the moment and it would be much safer to wait a bit longer to use it

dflems: Yeah, or make this a separate optional library

kmcbride: Alright, I've just pushed a commit to limit them to the latest OS versions.

gitrema (approved):

PR #220, opened 2022-05-31
Don't use enumerateByteRangesUsingBlock

gitrema:
This is basically a revert for
https://github.com/spotify/SPTDataLoader/commit/844112c2a56adcf5889af91b49ebd1ad21befe17

It is actually not clear how enumerateByteRangesUsingBlock can fix any
crash there but on the other hand is causing unnecessary callback and
delay in data delivering.


gitrema:
@kmcbride @dflems would you mind to take a look at it?

gitrema:
thanks for the link I missed that recommendation. On the other hand if I get it correctly they are talking about when the `bytes` property on NSData is used and that is when you need to have access to the underlying memory directly. The method implementation does not have that requirement and either append the coming data in an NSMutableData object or pass them to the observer

kmcbride:
OK, good point — it should be left up to the delegate to enumerate the bytes if needed.

kmcbride (approved):

PR #223, opened 2022-08-22
Allow performing requests for URLs that do not contain a host

kmcbride:
Certain URL types [supported](https://developer.apple.com/documentation/foundation/nsurlsession?language=objc#2934752) by `NSURLSession` are valid even though they do not contain a `host` property. Previously, the data loader would not even attempt to request these URLs. With this change it is possible to load local files as well as data URIs.

### DIFF
--- a/Sources/SPTDataLoader/SPTDataLoaderService.m
+++ b/Sources/SPTDataLoader/SPTDataLoaderService.m
@@ -207,27 +207,25 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)performRequest:(SPTDataLoaderRequest *)request
 requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
 {
-    if (request.cancellationToken.cancelled) {
+    if (request.URL == nil || request.cancellationToken.cancelled) {
         return;
     }
-    
-    if (request.URL.host == nil) {
-        return;
-    }
-    
-    NSString *host = [self.resolver addressForHost:(NSString * _Nonnull)request.URL.host];
-    NSString *requestHost = request.URL.host;
-    if (![host isEqualToString:requestHost] && host) {
-        NSURLComponents *requestComponents = [NSURLComponents componentsWithURL:request.URL resolvingAgainstBaseURL:NO];
-        requestComponents.host = host;
-        
-        NSURL *URL = requestComponents.URL;
-        
-        if (URL == nil) {
-            return;
+
+    if (request.URL.host != nil && self.resolver != nil) {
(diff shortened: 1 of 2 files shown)


dflems (approved):

gitrema:
SPTDataLoader was written with the intention to be a network library and more specific to handle HTTP requests and being implemented on top of URLSession to reduce its binary footprint. Allowing requests that are not meant to be HTTP to go through, it means to expand the surface in which the library operates so I would have expected to see an improvement in error reporting instead of allowing different types of URLs to go through.

kmcbride:
@gitrema fair point, but as far as I can tell nowhere in the library has it ever been enforced that requests or responses be limited to HTTP(S) — if that is the intention then I'm surprised by the lack of doing anything to prevent it. It seems to me that the use case addressed in this PR was originally supported and is now broken (probably unintentionally) because of a change to nullability handling: [host resolution introduction](https://github.com/spotify/SPTDataLoader/commit/e48c6fc341cf5b892d5231570c4ace334e8b6c8b#diff-091ce60264f6a4ba60ca7141559d0e914140c764fbf4086cca2f72e0ecd22c0dR80-R86) → [resolved host nil check](https://github.com/spotify/SPTDataLoader/commit/1377364b01a30745165c1e8f13d1627a271139c9#diff-091ce60264f6a4ba60ca7141559d0e914140c764fbf4086cca2f72e0ecd22c0dR82) → [nullability refactor](https://github.com/spotify/SPTDataLoader/commit/a1f2c56465c5a18e4ab12dee93d451f042bab297#diff-091ce60264f6a4ba60ca7141559d0e914140c764fbf4086cca2f72e0ecd22c0dR145-R151)

gitrema:
Thanks for pointing out the history by sharing previous PRs. I cannot find any mention in the README about usage of SPTDataLoader for something else other than networking. It does not mean that cannot be used to load other type of URLs of course even though it might not have been the intended use case. With that said if this PR is meant as bug-fix of a previously allowed behaviour it makes sense to merge it even though I would have personally preferred to take the opportunity to restrict requests to network protocols only.

kmcbride:
I will go ahead and merge this to restore the functionality. If it's decided that request scheme restrictions should be enforced, it can be done further up the call stack (eg: [here](https://github.com/spotify/SPTDataLoader/blob/2.2.0/Sources/SPTDataLoader/SPTDataLoader.m#L93-L127)).

kmcbride:
@gitrema I just noticed that [this portion](https://github.com/spotify/SPTDataLoader/blob/2.2.0/README.md?plain=1#L60) of the readme specifically calls out support for protocols beyond HTTP.

PR #226, opened 2022-11-04
Swift: Add decodable response type parameter

kmcbride:
Follows the common convention of including a type parameter, allowing the API call site to read a little nicer in cases where the response type cannot be inferred (or needs to differ from what's inferred).

Example:
```swift
struct SomeDecodable: Decodable {
  let value: Bool
}

// Before
responseDecodable { (response: Response<SomeDecodable, Error>) in
  methodAcceptingOptionalBool(response.result.success?.value)
}

// After
responseDecodable(type: SomeDecodable.self) { response in
  methodAcceptingOptionalBool(response.result.success?.value)
}
```


dflems (approved):
LGTM

PR #227, opened 2022-12-21
Add shouldStopRedirection to SPTDataLoaderRequest

im4nkur:
### What has changed
- Add a bool `shouldStopRedirection` to `SPTDataLoaderRequest`

### Why this was changed
- Currently url redirection happens automatically when using SPTDataLoader. To give user more control over the http url redirection a bool `shouldStopRedirection` is added to `SPTDataLoaderRequest`.


Sources/SPTDataLoader/SPTDataLoaderRequestTaskHandler.m
```diff
@@ -179,7 +181,7 @@ - (NSURLSessionResponseDisposition)receiveResponse:(NSURLResponse *)response
 - (BOOL)mayRedirect
```
dflems, line 181: this method name doesn't imply that it mutates anything, but it bumps `redirectCount`. out of scope for this PR to change it though, just thought i'd bring it up

dflems (approved):

PR #228, opened 2023-02-20
Add optional crash reporter

gabrielzanoni:
### What has changed
- Adds optional `crashReporter` of type `SPTDataLoaderCrashReporter` to constructor

### Why this was changed
This is an attempt to improve crash logging. It takes into account that the app probably has multiple instances of and `SPTDataLoader` during its lifecycle, which is why the breadcrumbs are left only in critical spots:
1. When loads are cancelled
2. When session gets invalidated
3. When there's an attempt to make a request in a invalidated session

### DIFF
--- a/Sources/SPTDataLoader/SPTDataLoaderService.m
+++ b/Sources/SPTDataLoader/SPTDataLoaderService.m
@@ -16,6 +16,7 @@
 
 #import "SPTDataLoaderService+Private.h"
 
+#import <SPTDataLoader/SPTDataLoaderCrashReporter.h>
 #import <SPTDataLoader/SPTDataLoaderCancellationToken.h>
 #import <SPTDataLoader/SPTDataLoaderRateLimiter.h>
 #import <SPTDataLoader/SPTDataLoaderResolver.h>
@@ -41,9 +42,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong) NSOperationQueue *sessionQueue;
 @property (nonatomic, strong) NSMutableArray<SPTDataLoaderRequestTaskHandler *> *handlers;
 @property (nonatomic, strong) NSMapTable<id<SPTDataLoaderConsumptionObserver>, dispatch_queue_t> *consumptionObservers;
+@property (nonatomic, strong, nullable, readonly) id<SPTDataLoaderCrashReporter> crashReporter;
 @property (nonatomic, strong) SPTDataLoaderServerTrustPolicy *serverTrustPolicy;
 @property (nonatomic, weak, nullable) NSFileManager *fileManager;
 @property (nonatomic, weak, nullable) Class dataClass;
+@property (nonatomic, assign) BOOL isSessionInvalidated;
 
 @end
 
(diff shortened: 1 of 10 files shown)


Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -378,12 +468,17 @@ - (void)URLSession:(NSURLSession *)session
...
-    
+
     completionHandler(disposition, credential);
 }
 
 #pragma mark NSURLSessionTaskDelegate
 
+- (void)URLSession:(NSURLSession *)session didBecomeInvalidWithError:(nullable NSError *)error
+{
+    [_invalidatedSessions addObject:session];
```
originalWho, line 479: As an option, we can keep track of invalidated sessions without extra set: a Service owns a SessionSelector, which in turn manages only two (or rather, up to two) session instances, so I guess we could leverage that instead.
Although I'm wondering if a Service should even start performing a request if its only Session has been invalidated. Wdyt?

gabrielzanoni: Not sure if I'm following why that would be better. This DataLoaderService is the one implementing `NSURLSessionTaskDelegate`, so if we were to go on that path this service would still have to tell SessionSelector which session is invalidated and then retrieve it after. It seems an unnecessary indirection 🤔 .

About not making a request on an invalidated session, Gianni made a change recently that prevents that from happening on a retry - but we'd still like to know if this happens in a regular request (i.e. leave breadcrumb and crash)

Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -41,9 +42,11 @@ @interface SPTDataLoaderService () <SPTDataLoaderRequestResponseHandlerDelegate,
 @property (nonatomic, strong) NSOperationQueue *sessionQueue;
 @property (nonatomic, strong) NSMutableArray<SPTDataLoaderRequestTaskHandler *> *handlers;
 @property (nonatomic, strong) NSMapTable<id<SPTDataLoaderConsumptionObserver>, dispatch_queue_t> *consumptionObservers;
+@property (nonatomic, strong, nullable, readonly) id<SPTDataLoaderCrashReporter> crashReporter;
 @property (nonatomic, strong) SPTDataLoaderServerTrustPolicy *serverTrustPolicy;
 @property (nonatomic, weak, nullable) NSFileManager *fileManager;
 @property (nonatomic, weak, nullable) Class dataClass;
+@property (nonatomic, strong) NSMutableSet<NSURLSession *> *invalidatedSessions;
```
gitrema, line 49: I would avoid to add this complexity in keeping track of what session got invalidated.
The `invalidateAndCancel` method always invalidates all the sessions handled by the SessionSelector.
My suggestion is to keep only a boolean to keep track of the session invalidation. This will simplify this PR

Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -405,16 +500,21 @@ - (void)URLSession:(NSURLSession *)session
         } else {
             handler.task = [session dataTaskWithRequest:handler.request.urlRequest];
         }
+    } else {
+        [_crashReporter leaveBreadcrumb:
+         [NSString stringWithFormat:@"Trying to retry on invalidated sesion. Policy: %ld, Error: %@",
+          (long)handler.request.backgroundPolicy, error.description]];
     }
```
gitrema, line 507: I am not sure about this one. The retry is an internal feature of SPTDataLoader. The `if` at line 402/497 was added to fix a bug where SPTDataLoader was trying to retry on an already invalidated sessions. And to add more on it, entering the `else` part does not guarantee the session was invalidated, it can just be that single task got cancelled and thus the message could be mis-leading as well

gabrielzanoni: Since this code can't retry on invalidated session anymore, I'll just remove this log then

originalWho (approved):

PR #229, opened 2023-02-20
Add a safeguard on NSURLSession invalidation

gitrema:
On a session invalidation, tasks are usually requested to be cancelled. the `didCompleteWithError` attempts always to create a new task even if the task had been cancelled. That is usally not an issue unless the session has been invalidated. In that case a NSGenericException is raied and the app most likely crashes.

### DIFF
--- a/Sources/SPTDataLoader/SPTDataLoaderService.m
+++ b/Sources/SPTDataLoader/SPTDataLoaderService.m
@@ -392,10 +392,19 @@ didCompleteWithError:(nullable NSError *)error
     if (handler == nil) {
         return;
     }
-    if (handler.request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
-        handler.task = [session downloadTaskWithRequest:handler.request.urlRequest];
-    } else {
-        handler.task = [session dataTaskWithRequest:handler.request.urlRequest];
+
+    //
+    // This is to avoid creating a new task on an already invalidated session
+    // when they are invalidated by `invalidateAndCancel` method.
+    // The check does not hurt even if the task is simply cancelled as there is no point
+    // in creating a new task that is not going to be used anyway.
+    //
+    if (!([error.domain isEqualToString:NSURLErrorDomain] && error.code == NSURLErrorCancelled)) {
+        if (handler.request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
+            handler.task = [session downloadTaskWithRequest:handler.request.urlRequest];
+        } else {
+            handler.task = [session dataTaskWithRequest:handler.request.urlRequest];
+        }
     }
     SPTDataLoaderResponse *response = [handler completeWithError:error];
     if (response == nil && !handler.cancelled) {
--- a/Tests/SPTDataLoader/SPTDataLoaderServiceTest.m
+++ b/Tests/SPTDataLoader/SPTDataLoaderServiceTest.m
@@ -719,6 +719,24 @@
     XCTAssertEqualObjects(handler.task, task);
 }
 
+- (void)testDoNotRecreateTaskWhenTaskIsCancelled
+{
+    SPTDataLoaderRequestResponseHandlerMock *requestResponseHandlerMock = [SPTDataLoaderRequestResponseHandlerMock new];
+    NSURL *URL = [NSURL URLWithString:@"https://localhost"];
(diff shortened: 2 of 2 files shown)


Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -392,10 +392,17 @@ - (void)URLSession:(NSURLSession *)session
...
-    if (handler.request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
-        handler.task = [session downloadTaskWithRequest:handler.request.urlRequest];
-    } else {
-        handler.task = [session dataTaskWithRequest:handler.request.urlRequest];
+
+    //
+    // This is to avoid creating a new task on an already invalidated session
+    // Usually on session invalidation all tasks are requested to be cancelled.
+    //
+    if (!([error.domain isEqualToString:NSURLErrorDomain] && error.code == NSURLErrorCancelled)) {
```
gabrielzanoni, line 400: Wouldn't be safer to use the [URLSession:didBecomeInvalidWithError:](https://developer.apple.com/documentation/foundation/nsurlsessiondelegate/1407776-urlsession) delegate to check if the session has been invalidated?

gitrema:
@kmcbride Does this look reasonable to you?

Sources/SPTDataLoader/SPTDataLoaderService.m
gitrema: I don't think that code is not safe. If a task is cancelled there is really no point of creating a new task that won't be used anyway. So it is either an optimization or a safeguard.
We can look into [URLSession:didBecomeInvalidWithError:](https://developer.apple.com/documentation/foundation/nsurlsessiondelegate/1407776-urlsession) as a mean to reject new requests in a follow-up PR.

gabrielzanoni (approved):

Sources/SPTDataLoader/SPTDataLoaderService.m
gabrielzanoni: the "safe" part I mentioned was related to the comment `Usually on session invalidation all tasks are requested to be cancelled` - Is it `usually` or 100% of the times?

gitrema: There are two ways a session can be invalidated. Through `finishTasksAndInvalidate` or `invalidateAndCancel` in the former, the code I changed is not impacted. The NSURLSession instance will wait until all task are done without cancelling them in the latter instead all tasks are immediately cancelled.
I can update the comment to better clarify when the safeguard will do the job

gitrema: I have update the comment.

gabrielzanoni (approved):

Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -392,10 +392,19 @@ - (void)URLSession:(NSURLSession *)session
...
+        if (handler.request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
+            handler.task = [session downloadTaskWithRequest:handler.request.urlRequest];
+        } else {
+            handler.task = [session dataTaskWithRequest:handler.request.urlRequest];
+        }
     }
     SPTDataLoaderResponse *response = [handler completeWithError:error];
     if (response == nil && !handler.cancelled) {
```
kmcbride, line 410: Seems like a reasonable change overall. I _think_ you could leverage the [existing error check](https://github.com/spotify/SPTDataLoader/blob/b0c37dc5ed258dc88ddad56fb2f3ba7fa224ace6/Sources/SPTDataLoader/SPTDataLoaderRequestTaskHandler.m#L120-L124) while keeping the same behavior by moving the task re-creation into the conditional. eg:
```suggestion

    SPTDataLoaderResponse *response = [handler completeWithError:error];
    if (response == nil && !handler.cancelled) {
        if (handler.request.backgroundPolicy == SPTDataLoaderRequestBackgroundPolicyAlways) {
            handler.task = [session downloadTaskWithRequest:handler.request.urlRequest];
        } else {
            handler.task = [session dataTaskWithRequest:handler.request.urlRequest];
        }

```

kmcbride (approved):
Seems OK. Could be nice to refactor in some way where the task is only re-created if a retry will be attempted.

PR #233, opened 2023-05-04
Prevent a task to be created from an invalid session

gitrema:
Authoriser are not cancellable and so after calling `invalidateAndCancel` method on the SPTDataLoaderService some in-flight request still going through authoriser might end up to create a URLSessionTask on an already invalidated one with the result of a throwing exception

### DIFF
--- a/Sources/SPTDataLoader/SPTDataLoaderService.m
+++ b/Sources/SPTDataLoader/SPTDataLoaderService.m
@@ -49,6 +49,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong) SPTDataLoaderServerTrustPolicy *serverTrustPolicy;
 @property (nonatomic, weak, nullable) NSFileManager *fileManager;
 @property (nonatomic, weak, nullable) Class dataClass;
+@property (nonatomic, assign) BOOL sessionInvalidated;
 
 @end
 
@@ -138,6 +139,7 @@ NS_ASSUME_NONNULL_BEGIN
 
         _fileManager = [NSFileManager defaultManager];
         _dataClass = [NSData class];
+        _sessionInvalidated = NO;
     }
 
     return self;
@@ -224,6 +226,10 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)performRequest:(SPTDataLoaderRequest *)request
 requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
 {
+    if (self.sessionInvalidated) {
+        [requestResponseHandler cancelledRequest:request];
+        return;
+    }
     if (request.URL == nil || request.cancellationToken.cancelled) {
         return;
     }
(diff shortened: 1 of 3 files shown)


Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -138,6 +139,7 @@ - (instancetype)initWithConfiguration:(NSURLSessionConfiguration *)configuration
 
         _fileManager = [NSFileManager defaultManager];
         _dataClass = [NSData class];
+        _sessionInvalidated = NO;
```
originalWho, line 142: nit: this is redundant as it's `NO` by default

Sources/SPTDataLoader/SPTDataLoaderService.m
```diff
@@ -224,6 +226,10 @@ - (NSURLSessionTask *)createTaskForRequest:(SPTDataLoaderRequest *)request
 - (void)performRequest:(SPTDataLoaderRequest *)request
 requestResponseHandler:(id<SPTDataLoaderRequestResponseHandler>)requestResponseHandler
 {
+    if (self.sessionInvalidated) {
```
originalWho, line 229: Maybe consider moving this check to `createTaskForRequest:`? Otherwise we're risking to face the same issue via the `requestTaskHandlerNeedsNewTask:` method

Tests/SPTDataLoader/SPTDataLoaderServiceTest.m
```diff
@@ -767,4 +767,14 @@ - (void)testRequestingDataURL
...
 }
 
+- (void)testCancellingRequestOnSessionInvalidation
+{
+    SPTDataLoaderRequestResponseHandlerMock *requestResponseHandlerMock = [SPTDataLoaderRequestResponseHandlerMock new];
+    NSURL *URL = [NSURL URLWithString:@"https://localhost"];
+    SPTDataLoaderRequest *request = [SPTDataLoaderRequest requestWithURL:URL sourceIdentifier:@""];
+    [self.service invalidateAndCancel];
+    [self.service requestResponseHandler:requestResponseHandlerMock performRequest:request];
+    XCTAssertEqual(requestResponseHandlerMock.numberOfCancelledRequestCalls, 1u);
```
originalWho, line 777:
```suggestion
    XCTAssertEqual(requestResponseHandlerMock.numberOfCancelledRequestCalls, 1);
```

originalWho (approved):

Sources/SPTDataLoader/SPTDataLoaderService.m
gitrema: I like to be explicit, if that is not an issue to have it there

Sources/SPTDataLoader/SPTDataLoaderService.m
gitrema: I considered it but then how can `cancelledRequest` be called easily? I should change the signature of `createTaskForRequest` and return some nullable value and check for it in all call sites? To me `createTaskForRequest` is not designed to fail and when called URLSession should be valid and this is exactly the bug I am trying to fix here

originalWho: Ah I see your point now, thanks for clarifying!

